PR #6, opened 2025-05-20
test: Add integration test with KinD

minhtule:
## Changes

- Add an integration test with [a KinD cluster](http://kind.sigs.k8s.io)
  + Test that Twingate identity is forwarded correctly to k8s API server
- Add code coverage and CI for integration test


internal/connect/connect.go
```diff
@@ -100,7 +101,8 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (claims *
 	if !strings.EqualFold(host, gatClaims.Resource.Address) {
 		response = unauthorizedResponse
 
-		return nil, response, fmt.Errorf("failed to verify CONNECT destination: %s(err: %w) with token resource address %s", host, err, gatClaims.Resource.Address)
```
minhtule, line 103: `err` is nil here so we don't need to include it.

Makefile
```diff
@@ -24,6 +24,12 @@ HELP_FUN = \
     }; \
     print "\n"; }
 
+COVPROFILE_UNIT := covprofile-unit
+COVPROFILE_INTEGRATION := covprofile-integration
+COVERED_PACKAGES := ./cmd/...,./internal/...
```
minhtule, line 29: Go test coverage doesn't seem to have an option to exclude some directories (like `test/fake/` package) so I explicitly specify which packages to be included for coverage.

minhtule:
Closing this in favor of https://github.com/Twingate/kubernetes-access-gateway/pull/13

PR #8, opened 2025-05-20
fix: Make `tls` object optional in Helm JSON schema

minhtule:
## Changes

`tls` object has a default behavior to auto-generated TLS certificates so the user doesn't need to specify it in the Helm value file

### DIFF
--- a/deploy/gateway/values.schema.json
+++ b/deploy/gateway/values.schema.json
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/schema#",
   "type": "object",
   "title": "JSON schema for Twingate Gateway Helm Chart values",
-  "required": ["twingate", "tls"],
+  "required": ["twingate"],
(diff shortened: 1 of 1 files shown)


deploy/gateway/values.schema.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/schema#",
   "type": "object",
   "title": "JSON schema for Twingate Gateway Helm Chart values",
-  "required": ["twingate", "tls"],
```
ekampf, line 5: need to also add defaults in template?

minhtule: Double-checked that this change was not needed. When the user of the Helm chart doesn't specify `tls` object, the default values would be used. Helm lint doesn't fail either. The error `Schema validation: Missing required property 'tls'` only appears in certain editor...

PR #9, opened 2025-05-21
refactor: Cleaner configuration setup using Viper

minhtule:
## Changes

- Use Viper to bind all command flags so they can be set as environment variables.
- Standardize flag names

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -80,7 +80,7 @@ func start(newProxy ProxyFactory) error {
 	return nil
 }
 
-func init() {
+func init() { //nolint:gochecknoinits
 	viper.SetEnvPrefix("TWINGATE")
 	viper.AutomaticEnv()
 
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -96,7 +96,7 @@ func init() { //nolint:gochecknoinits
 	flags.String("tlsKey", "", "Path to the TLS key for the Gateway")
 
 	// Kubernetes flags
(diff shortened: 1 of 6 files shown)


cmd/start.go
```diff
@@ -87,25 +80,30 @@ func start(newProxy ProxyFactory) error {
...
+
+	// Kubernetes flags
+	flags.String("k8sAPIServerCA", "", "Path to the K8sAPIServerCA certificate for the Kubernetes API server")
+	flags.String("k8sAPIServerToken", "", "Bearer token to authenticate to the Kubernetes API server")
+
+	// Misc flags
+	flags.BoolP("debug", "d", false, "Run in debug mode")
+
+	if err := viper.BindPFlags(flags); err != nil {
+		panic(fmt.Sprintf("failed to bind flags: %v", err))
```
minhtule, line 106: - Rename `ca` flag to `k8sAPIServerCA`.
- Add `port` flag so it's configurable and passed to the proxy's config.
- Reorder flags into related groups.

internal/httpproxy/http_proxy.go
```diff
@@ -45,13 +45,12 @@ const (
 )
 
 type Config struct {
-	CA                string
 	TLSCert           string
 	TLSKey            string
 	K8sAPIServerToken string
-	K8sAPIServerURL   string
```
minhtule, line 52: We don't use this field anymore (we use the resource's address in GAT instead) but I forgot to remove this.

sghiocel (approved):

PR #13, opened 2025-05-30
test: Add integration test with KinD

minhtule:
## Changes

- Add an integration test with [a KinD cluster](http://kind.sigs.k8s.io/)
  + Test that Twingate identity is forwarded correctly to k8s API server
  + Test Gateway logs the REST request and session recording
- Create fake Controller and Client that could be used in testing and local development
- Add code coverage and CI for integration test


Makefile
```diff
@@ -24,6 +24,12 @@ HELP_FUN = \
     }; \
     print "\n"; }
 
+COVPROFILE_UNIT := covprofile-unit
+COVPROFILE_INTEGRATION := covprofile-integration
+COVERED_PACKAGES := ./cmd/...,./internal/...
```
minhtule, line 29: Go test coverage doesn't seem to have an option to exclude some directories (like test/fake/ package) so I explicitly specify which packages to be included for coverage.

test/fake/client.go
```diff
@@ -0,0 +1,229 @@
...
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS13,
+		ServerName: "127.0.0.1",
+		RootCAs:    caCertPool,
+	}
+
+	// Manually create a TCP connection
+	conn, err := net.Dial("tcp", c.proxyAddress)
+	if err != nil {
+		c.logger.Error("Failed to connect to proxy", zap.Error(err))
```
sghiocel, line 129: missing `return` here?

test/fake/client.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	defer conn.Close()
+
+	// Enable outer TLS (between Twingate client and proxy)
+	proxyTLSConn := tls.Client(conn, tlsConfig)
+	if err := proxyTLSConn.Handshake(); err != nil {
+		c.logger.Error("TLS handshake failed(proxy)", zap.Error(err))
```
sghiocel, line 136: `return` here on error?

test/fake/client.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
```
sghiocel, line 138: should also add `defer proxyTLSConn.close()` here to ensure clean TLS shutdown with `close-notify`

sghiocel (approved):

PR #16, opened 2025-06-02
fix: Fix the default `host` value to `twingate.com` in Helm chart

minhtule:
## Changes

Fix the default `host` value to "twingate.com" in Helm chart


romankor (approved):
GTM

PR #24, opened 2025-06-09
feat: Support bearer token refresh for in-cluster deployment

minhtule:
## Changes

Use client-go's `NewBearerAuthWithRefreshRoundTripper` transport to refresh the bearer token from the provided file (`/var/run/secrets/kubernetes.io/serviceaccount/token` for in-cluster deployment)

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -64,11 +64,12 @@ func start(newProxy ProxyFactory) error {
 		},
 	}
 
-	if inClusterK8sCfg, _ := rest.InClusterConfig(); inClusterK8sCfg != nil {
+	if inClusterK8sCfg, err := rest.InClusterConfig(); inClusterK8sCfg != nil {
 		logger.Info("Using in-cluster configuration")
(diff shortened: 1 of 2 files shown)


cmd/start.go
```diff
@@ -64,11 +64,12 @@ func start(newProxy ProxyFactory) error {
 		},
 	}
 
-	if inClusterK8sCfg, _ := rest.InClusterConfig(); inClusterK8sCfg != nil {
+	if inClusterK8sCfg, err := rest.InClusterConfig(); inClusterK8sCfg != nil {
```
minhtule, line 67: Not related this feature but this fixes an issue that log in the else-if block in [line 76](https://github.com/Twingate/kubernetes-access-gateway/pull/24/files#diff-e3ad76770e5ac8ed654e8c5585782469e67e0d981adb9153604109263f22e93eR76) is always printed when the gateway is running outside the k8s cluster.

PR #29, opened 2025-06-16
test: Add an integration test for concurrent users

minhtule:
## Changes

- Add an integration test for concurrent users accessing the Gateway
- Extract helper structs and functions to `integration/testutil` package
  + Simplify client `Kubectl` setup. We can pass the context information via `kubectl` CLI options e.g.
    ```
    kubectl --server {server-url} --certificate-authority {ca-file-path} --token void ...
    ```
    so there is no need to update the `~/.kube/config` file (and clean up at the end of the test!)

### DIFF
--- a/test/integration/concurrent_users_test.go
+++ b/test/integration/concurrent_users_test.go
@@ -188,7 +188,6 @@ func TestConcurrentUsers(t *testing.T) {
 				lastLogTime := time.Time{}
 
 				output, err := user.Kubectl.Command(cmd.args...)
-
 				if err != nil {
 					t.Logf("Failed to run kubectl %s: %v", cmd.name, err)
 				}


test/integration/authentication_test.go
```diff
@@ -58,6 +41,8 @@ func TestKubernetesAuthentication(t *testing.T) {
 		rootCmd := cmd.GetRootCommand()
 		rootCmd.SetArgs([]string{
 			"start",
+			"--port",
+			strconv.Itoa(gatewayPort),
```
minhtule, line 45: Ideally we should be able to set `port=0` like `httptest.Server` to pick a random port to avoid port collision. Currently we can't do that because we don't have a way to retrieve that random port value. Hence, I hardcore the port value (`8443` and `8444`) for these tests.

test/integration/authentication_test.go
```diff
@@ -94,327 +79,45 @@ func TestKubernetesAuthentication(t *testing.T) {
...
-  name: gateway-test-user
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: edit
-subjects:
-- kind: User
-  name: "alex@acme.com"
-  apiGroup: rbac.authorization.k8s.io
-`
```
minhtule, line 205: This cluster role binding YAML is extracted to a new method [`testutil.CreateK8sRoleBinding`](https://github.com/Twingate/kubernetes-access-gateway/pull/29/files#diff-eed45dbc0ee3e34a23d4a90d5b9cfb09a74e06a8a034c0f5f6e4c529df4f1b0dR160) so we could bind the `edit` role to more than 1 user.

test/integration/testutil/assertions.go
```diff
@@ -0,0 +1,79 @@
...
+	t.Helper()
+
+	var podList corev1.PodList
+	if err := json.Unmarshal(output, &podList); err != nil {
+		t.Fatalf("Failed to parse kubectl get pods output: %v", err)
+	}
+
+	assert.Len(t, podList.Items, 1)
+	assert.Equal(t, "test-pod", podList.Items[0].Name)
+}
```
minhtule, line 40: This assertion is new. The rest are extracted from `authentication_test.go`

test/integration/concurrent_users_test.go
```diff
@@ -0,0 +1,234 @@
...
+					},
+				},
+			}
+			rand.Shuffle(len(commands), func(i, j int) {
+				commands[i], commands[j] = commands[j], commands[i]
+			})
+
+			t.Logf("User %s is running commands: %s", user.ID, commands)
+
+			lastLogTime := time.Time{}
```
sghiocel, line 187: can this be inside the `for` loop, before running the command?

test/integration/concurrent_users_test.go
```diff
@@ -0,0 +1,234 @@
...
+			"--k8sAPIServerToken",
+			kindBearerToken,
+			"--fakeControllerURL",
+			controller.URL,
+			"--metricsPort",
+			"0",
+		})
+
+		err := rootCmd.Execute()
+		t.Logf("Failed to start Gateway: %v", err)
```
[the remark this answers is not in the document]
minhtule: I think so. We only get here if the Gateway port `8444` is already taken

test/integration/concurrent_users_test.go
minhtule: Yes, I think it's nicer! We can remove [line 215](https://github.com/Twingate/kubernetes-access-gateway/pull/29/files#diff-786059d243e2e03a9358891e734671f3d86a85cfa09464b8b785ff1ca9735e43R215) that updates this `lastLogTime` at the end of each command.

PR #31, opened 2025-06-16
feat: Add Prometheus Go client and Go process metrics

clement0010:
## Changes
- Add Prometheus Go Client [library](https://github.com/prometheus/client_golang) 
- Update Helm chart to enable exposing Prometheus metrics on a separate port (`9090`)
- Allow setting `TWINGATE_PORT` and `TWINGATE_METRICSPORT` via Helm `values.yaml`

## Notes
You can test this by running a simple Prometheus server:
1. Install prometheus server: `brew install prometheus`
2. Save the following config file as `prometheus.yml`
```yml
global:
  scrape_interval: 5s
  evaluation_interval: 5s

scrape_configs:
  - job_name: 'metrics-test-server'
    static_configs:
      - targets: ['localhost:9090']
```
3. Forward the Prometheus port to local machine: `kubectl port-forward svc/<gateway service> 9090:9090`
4. Run `prometheus --config.file=prometheus.yml --web.listen-address=:9091`
5. Access the Prometheus dashboard at `localhost:9091`

## Reference
- Grafana's Prometheus: https://github.com/grafana/grafana/blob/main/pkg/infra/metrics/metrics.go


deploy/gateway/templates/service.yaml
```diff
@@ -23,5 +23,9 @@ spec:
       port: 443
       targetPort: https
       protocol: TCP
+    - name: metrics
+      port: 9090
+      targetPort: metrics
+      protocol: TCP
```
minhtule, line 29: We shouldn't expose the metrics port in the service because that might expose Prometheus metrics externally. Prometheus could be configured to scrape metrics directly from the Gateway pods.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"k8sgateway/internal/version"
+)
+
+const Namespace = "twingate_gateway"
+
+type Config struct {
+	Port     string
+	Logger   *zap.SugaredLogger
+	Registry *prometheus.Registry
```
clement0010, line 22: We can't make these variables private because it's exported and used [here](https://github.com/Twingate/kubernetes-access-gateway/pull/31/files#diff-e3ad76770e5ac8ed654e8c5585782469e67e0d981adb9153604109263f22e93eR86-R88)

internal/metrics/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+
+		Addr:    fmt.Sprintf(":%v", config.Port),
+		Handler: mux,
+	}
+
+	logger.Infof("Starting metrics server on: %v", config.Port)
+
+	if err := server.ListenAndServe(); err != nil {
+		logger.Fatalf("Failed to start metrics server: %v", err)
+	}
```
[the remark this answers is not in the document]
clement0010: Hmmm, interesting. I am not sure if we want to continue running the application when the metrics server fails to start. @minhtule thoughts on this?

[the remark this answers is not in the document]
clement0010: Chatted on this offline, we should avoid using `logger.Fatalf` because it terminates the entire program. A better approach is to use `logger.Errorf` and return the error.

cmd/start.go
```diff
@@ -79,6 +80,17 @@ func start(newProxy ProxyFactory) error {
...
 
+	metricsPort := viper.GetString("metricsPort")
+	go func() {
+		err := metrics.Start(metrics.Config{
+			Port:   metricsPort,
+			Logger: zap.S(),
+		})
+		if err != nil {
+			logger.Errorf("failed to start metrics server: %v", err)
+		}
```
minhtule, line 91: - If the metrics server fails to start (e.g. port is not available to bind), this would silently fail. We should call `logger.Fatal` here so the program should stop immediately. The better way is to use [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) to setup both app and metrics server (and also support graceful shutdown) but we'll do another PR for that.
-  We should use the `zap.Error()` like [this](https://github.com/Twingate/kubernetes-access-gateway/blob/14873853c4cc97ecc24e96699d82899b237931b6/internal/httpproxy/http_proxy.go#L195) instead of embedding the error in the log message.

minhtule: Yes, I think metrics collection is as critical. It's better to stop the gateway than running without Telemetry.

minhtule (approved):
LGTM! 👍

PR #32, opened 2025-06-18
fix: improper websocket parsing

sghiocel:
## Changes
- Refactor and improve websocket parsing code, including fixing the current issues with double messages
- Improve testing for websocket frame and message parsing

### DIFF
--- a/internal/wsproxy/conn_test.go
+++ b/internal/wsproxy/conn_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/util/remotecommand"
 )
 
 type mockConn struct {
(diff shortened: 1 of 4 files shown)


internal/wsproxy/conn_test.go
```diff
@@ -82,47 +83,249 @@ func (m *mockRecorder) IsHeaderWritten() bool {
 	return m.headerWritten
 }
 
+type WebSocketMessage struct {
+	Bytes   []byte // raw bytes of the websocket frame
+	Payload []byte // the payload of the websocket frame
+}
```
minhtule, line 89: Perhaps we could avoid duplicating the payload with something like
```go
type WebSocketMessage struct {
	Headers   []byte // raw bytes of the websocket frame
	Payload []byte // the payload of the websocket frame
}

func (m WebSocketMessage)Raw() {
	return append(m.Headers, m.Payload)
}
```
?

internal/wsproxy/conn_test.go
```diff
@@ -82,47 +83,249 @@ func (m *mockRecorder) IsHeaderWritten() bool {
...
+	Bytes: []byte{
+		// Close Control Frame with "done" payload: (0x88, 0x06, 0x03, 0xe8, 0x64, 0x6f, 0x6e, 0x65)
+		// 0x88: FIN=1 (final frame), Opcode=0x8 (Close Frame)
+		// 0x06: Mask=0 (not masked), Payload Length=6 bytes
+		// 0x03, 0xE8: Close Status Code (1000: Normal Closure)
+		// 0x64, 0x6f, 0x6e, 0x65: UTF-8 for "done"
+		0x88, 0x06, 0x03, 0xe8, 0x64, 0x6f, 0x6e, 0x65,
+		0x82, 0x19, 0x4, 0x7b, 0x22, 0x77, 0x69, 0x64, 0x74, 0x68, 0x22, 0x3a, 0x34, 0x30,
+		0x2c, 0x22, 0x68, 0x65, 0x69, 0x67, 0x68, 0x74, 0x22, 0x3a, 0x38, 0x30, 0x7d,
+	},
```
minhtule, line 114: Why doesn't CLOSE message have a `Payload`? 🤔

minhtule (approved):
LGTM! 👍

internal/wsproxy/conn_test.go
sghiocel: if the message contains a masked payload, then the raw bytes cannot be headers + payload appended, the raw bytes should have the masked payload and the payload we are asserting should be the unmasked final data

PR #36, opened 2025-06-19
feat: Add support for Prometheus `podMonitor`

clement0010:
## Changes
- Add `podmonitor.yaml` Helm template
- Update `values.yaml` to support configuring `podMonitor` for Prometheus

## Notes
Reference: https://github.com/bitnami/charts/blob/1f0d9e8d99735ea2479224b14cbf83a71be5e660/bitnami/redis/values.yaml#L2038-L2093


deploy/gateway/templates/podmonitor.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: {{ include "gateway.fullname" . }}
+  labels:
+    {{- include "gateway.labels" . | nindent 4 }}
+    {{- with .Values.metrics.podMonitor.additionalLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
```
minhtule, line 14: Do we really need this? Would Prometheus scrape the namespace the `PodMonitor` is deployed in?

deploy/gateway/tests/snapshot_test.yaml
```diff
@@ -5,5 +5,9 @@ excludeTemplates:
   - tls-secret.yaml
 tests:
   - it: should render
+    set:
+      metrics:
+        podMonitor:
+          enabled: true
```
minhtule, line 11: I think it's better to create a different test here to make sure we don't create `PodMonitor` by default.

deploy/gateway/templates/podmonitor.yaml
clement0010: Hmm, nope, this is just setting the default value explicitly. Prometheus would discover the pods in the same namespace as the `PodMonitor`

PR #41, opened 2025-06-20
refactor: Modernize tests to use `testify`

clement0010:
## Changes
- Update unit test to use `testify/assert` and `testify/require`

### DIFF
--- a/internal/httpproxy/http_proxy_test.go
+++ b/internal/httpproxy/http_proxy_test.go
@@ -118,15 +118,8 @@ func TestTruncateBody(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			result := truncateBody(tc.input)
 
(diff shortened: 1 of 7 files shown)


internal/httpproxy/http_proxy_test.go
```diff
@@ -568,47 +560,35 @@ func TestProxy_ForwardRequest(t *testing.T) {
 	// manually create a TCP connection to be able to reuse for
 	// HTTPS CONNECT (downstream proxy) then HTTPS requests (client)
 	conn, err := net.Dial("tcp", "127.0.0.1:45678")
-	if err != nil {
-		t.Fatalf("Failed to connect to proxy: %v", err)
-	}
+	require.NoError(t, err, "Failed to connect to proxy")
```
clement0010, line 563: Should we still keep these messages?

PR #44, opened 2025-06-24
feat: Add `metrics.podMonitor` to Helm `values.yaml` file JSON schema

clement0010:
## Changes
- Add `metrics.podMonitor` schema to `values.yaml` file JSON schema

## Notes
Adapted from: https://github.com/bitnami/charts/blob/0ec66e0712cea64de3afbc6e433664a2d8e9672c/bitnami/redis/values.schema.json#L2930-L3008

### DIFF
--- a/deploy/gateway/values.schema.json
+++ b/deploy/gateway/values.schema.json
@@ -94,11 +94,9 @@
     },
     "metrics": {
       "type": "object",
-      "required": ["podMonitor"],
       "properties": {
         "podMonitor": {
           "type": "object",
-          "required": ["enabled"],
           "properties": {
             "enabled": {
               "type": "boolean",


deploy/gateway/templates/podmonitor.yaml
```diff
@@ -14,7 +14,9 @@ spec:
       {{- include "gateway.selectorLabels" . | nindent 6 }}
   podMetricsEndpoints:
     - port: metrics
+      {{- if .Values.metrics.podMonitor.interval }}
```
clement0010, line 17: When `metrics.podMonitor.interval` is not provided, we shouldn't render an empty `interval` field in the template

minhtule (approved):

PR #51, opened 2025-07-02
feat: Add recorded session duration metrics

clement0010:
## Changes
- Add `twingate_gateway_recorded_session_duration_seconds` duration metric for recorded sessions. E.g. `kubectl exec -it`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,6 +24,7 @@ linters:
     - varnamelen
     - wrapcheck
     - noinlineerr
+    - wsl # Deprecated since v2.2.0 and replaced by wsl_v5
   exclusions:
     presets:
       - comments
--- a/test/integration/concurrent_users_test.go
+++ b/test/integration/concurrent_users_test.go
@@ -170,7 +170,9 @@ func TestConcurrentUsers(t *testing.T) {
 					args: []string{"exec", "test-pod", "--", "cat", "/etc/hostname"},
 					assertOutputFn: func(t *testing.T, output []byte) {
 						t.Helper()
-						assert.Equal(t, "test-pod\n", string(output))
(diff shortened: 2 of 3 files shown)


internal/wsproxy/recorder.go
```diff
@@ -21,6 +22,19 @@ var (
 	errAlreadyFinished     = errors.New("recording already finished")
 )
 
+var (
+	recordedSessionDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "twingate_gateway",
```
minhtule, line 27: When we change the `namespace` constant in `internal/metrics.go`, how do we know to change this string here?

internal/wsproxy/recorder.go
```diff
@@ -21,6 +22,24 @@ var (
...
+
+func RegisterRecordedSessionMetrics(namespace string, registry *prometheus.Registry) {
+	initializeMetricsOnce.Do(func() {
+		recordedSessionDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "recorded_session_duration_seconds",
+			Help:      "Duration of WebSocket session in seconds",
+			Buckets:   []float64{1, 5, 10, 30, 60, 300, 600, 1800, 3600},
+		})
+	})
```
clement0010, line 38: We detected a data race in the tests (see [CI](https://github.com/Twingate/kubernetes-access-gateway/actions/runs/16132786927/job/45523187155)), using `sync.Once` here should fix it.

I think it's fair to assume we won't initialize these metrics again with a different namespace. So this solution should be fine. I am open to other alternatives

sghiocel (approved):

PR #54, opened 2025-07-03
chore: Add license header to Go files

ekampf:
## Changes

Add `goheader` to golangci-lint with out license header


stuloh (approved):
trailing `//` in open source comment header not needed but inclusion is intentional

PR #60, opened 2025-07-08
chore: Fix `golang-lint` warning

clement0010:
## Changes
- Fix this warning message from `golang-lint` by disabling `wsl` linter as `wsl_v5` is enabled by default
```Go
WARN The linter 'wsl' is deprecated (since v2.2.0) due to: new major version. Replaced by wsl_v5. 
WARN [linter] `allow-multiline-assign` is deprecated and always allowed in wsl >= v5 
WARN Suggested new configuration:
linters:
  enable:
    - wsl_v5
  settings:
    wsl_v5:
      allow-first-in-block: true
      allow-whole-block: false
      branch-max-lines: 2
```
- Fix `unhandled-error` from `revive` linter


.golangci.yml
```diff
@@ -5,6 +5,8 @@ run:
 
 linters:
   default: all
+  enable:
+    - wsl_v5
```
minhtule, line 9: Do we need to enable this linter when we already enable all linters in the line above?

minhtule (changes requested):

ekampf:
This warning only shows on latest linter but CI doesnt use it...  I'd expect this PR to upgrade the linter version too?

minhtule (approved):
LGTM 👍

PR #65, opened 2025-07-10
test: Fix flaky integration test

minhtule:
## Changes

Fix flaky test due to warning message from `kubectl exec ... -- sleep 2` command

### DIFF
--- a/test/integration/concurrent_users_test.go
+++ b/test/integration/concurrent_users_test.go
@@ -158,25 +158,40 @@ func TestConcurrentUsers(t *testing.T) {
 					args: []string{"exec", "test-pod", "--", "sleep", "2"},
 					assertOutputFn: func(t *testing.T, output []byte) {
 						t.Helper()
-						assert.Empty(t, string(output))
+						// `kubectl exec` has sporadic warning message "Unknown stream id 1, discarding message", especially when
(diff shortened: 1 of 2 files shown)


test/integration/concurrent_users_test.go
```diff
@@ -158,25 +158,40 @@ func TestConcurrentUsers(t *testing.T) {
 					args: []string{"exec", "test-pod", "--", "sleep", "2"},
 					assertOutputFn: func(t *testing.T, output []byte) {
 						t.Helper()
-						assert.Empty(t, string(output))
+						// `kubectl exec` has sporadic warning message "Unknown stream id 1, discarding message", especially when
+						// there are multiple concurrent streams so while `sleep` command has no output, `assert.Empty` would occasionally
+						// fail. We'll assert the output doesn't contain the other `cat /etc/hostname` output instead.
+						assert.NotContains(t, string(output), testutil.TestPodName)
```
minhtule, line 164: I extracted `test-pod` to a constant because if we change the pod name in the setup but doesn't change it here, the test would always pass even when there is a bug that somehow the `cat /etc/hostname` output gets "leaked" to this command.

sghiocel (approved):

PR #70, opened 2025-07-15
fix: Use `ProxyConn.Close()` to close healthcheck TLS connection

clement0010:
## Changes
This is a follow-up PR for https://github.com/Twingate/kubernetes-access-gateway/pull/45, we should let `ProxyConn.Close()` to handle closing the TLS connection instead of closing it manually


minhtule (approved):
LGTM 👍

sghiocel (approved):

PR #71, opened 2025-07-15
feat: Add Client - Gateway TCP connection Prometheus metrics

clement0010:
## Changes
- Measure the connections between Client and Gateway TCP connection: `activeConnections`, `connectionDuration` and `connectionsTotal`
- Measure the total count (`connectTotal`) and duration (`connectDuration`) for HTTP Connect
- Create `proxyConnMetricsTracker` to handle measuring TCP connection and HTTP Connect metrics.

## Notes
When we run `kubectl`, `poll.DeadlineExceededError` occurs in `ProxyConn.Read()`, which is a non-fatal exception. As such, we won't be tracking `connectionError` for now since non-fatal exceptions are recorded.


internal/httpproxy/http_proxy.go
```diff
@@ -124,6 +124,9 @@ func (p *ProxyConn) authenticate() error {
 		return err
 	}
 
+	// Store reference to original metrics wrapper before TLS upgrade
+	originalMetricsConn, hasMetrics := p.Conn.(*connWithMetrics)
```
clement0010, line 128: The reason we need to store the `originalMetricsConn` is because in line 131, we upgrade the connection to TLS. 

The `tlsConnectConn.conn` has the `connWithMetrics` but we can't access unexported `tls.conn` field.

minhtule: Regardless of changing the underlying connection `p.Conn`, I think this type casting is not very nice because `ProxyConn` creates this `connWithMetrics` in the constructor. I would add an explicit field to reference the underlying TCP connection e.g.
```go
type ProxyConn struct {
    Conn net.Conn
    tcpConn *connWithMetrics
    // other fields...
}
```

internal/httpproxy/http_proxy_test.go
```diff
@@ -643,3 +646,33 @@ func TestShouldSkipWebSocketRequest(t *testing.T) {
...
+			expectedResult: true,
+		},
+		{
+			name:           "POST request to healthcheck path",
+			request:        httptest.NewRequest(http.MethodPost, healthCheckPath, nil),
+			expectedResult: false,
+		},
+		{
+			name:           "Proxy request",
+			request:        httptest.NewRequest(http.MethodGet, "/api/v1/namespaces/default/pods", nil),
```
minhtule, line 668: nit: It doesn't matter for this test but to be precise, a "proxy request" starts with a HTTP CONNECT, not GET (This GET request comes after the CONNECT request).

internal/httpproxy/http_proxy.go
```diff
@@ -146,8 +149,12 @@ func (p *ProxyConn) authenticate() error {
 		return err
 	}
 
+	if hasMetrics {
+		originalMetricsConn.setConnectionCategory(req)
```
minhtule, line 153: This method seems unnecessary coupling (why does a TCP connection need to know about HTTP request?). Would setting the metric label explicitly simpler here e.g.
```go
if isHealthCheckRequest(req) {
  originalMetricsConn.connectionCategory = connectionCategoryHealth
  // other code
}
originalMetricsConn.connectionCategory = connectionCategoryProxy
```
?

internal/httpproxy/http_proxy.go
clement0010: I agree that type assertion is not nice here too. If we add `tcpConn *connWithMetrics` for reference, we'd have both `Conn` and `tcpConn` referring to the same `Conn` initially ([code](https://github.com/Twingate/kubernetes-access-gateway/pull/71/files#diff-309da065b292d8850fba689660bc96fadda998d8562fd3ad8f1c6930cb30ac61R261-R262)). I think it's not the best either? But I don't have a better suggestion :s

minhtule: > I think it's not the best either?

Not ideal but I think it's the best we could do here 🙈

minhtule: Yes, that's a great idea! This `connWithMetrics` is pretty coupled with the `ProxyConn`.

clement0010: Yeah, I think this approach is nicer. One small thing to note is that we still need to implement type assertion [here](https://github.com/Twingate/kubernetes-access-gateway/pull/71/files#diff-309da065b292d8850fba689660bc96fadda998d8562fd3ad8f1c6930cb30ac61R380-R385). I suppose we cannot avoid type assertion regardless...

internal/httpproxy/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+type proxyConnMetrics struct {
+	activeConn   prometheus.Gauge
+	connTotal    *prometheus.CounterVec
+	connDuration *prometheus.HistogramVec
+
+	connectTotal    *prometheus.CounterVec
+	connectDuration *prometheus.HistogramVec
+}
+
+func registerProxyConnMetrics(registry *prometheus.Registry) *proxyConnMetrics {
```
clement0010, line 30: Previously, we declared the metrics as shared global variables (see [diff](https://github.com/Twingate/kubernetes-access-gateway/pull/71/commits/b26799334fc4fc4ec27ff41c45295b467bb4b606)), which resulted in data races in test cases (see [here](https://github.com/Twingate/kubernetes-access-gateway/actions/runs/16491431219/job/46626802915)). We'd have different test case registering and writing to the shared metrics. Thus, I moved the metrics here to ensure each test case has its own metrics instance.

internal/httpproxy/http_proxy_test.go
```diff
@@ -100,6 +101,16 @@ func TestProxyConn_Close(t *testing.T) {
 		assert.Fail(t, "Timer should have been stopped")
 	default:
 	}
+
+	// Ensure metrics are only measured once
+	_ = proxyConn.Close()
```
sghiocel, line 106: Why call `Close()` again here, it's already closed? Also, below is doing:

```
	proxyConn.once.Do(func() {
 		called = true
 	})
 ```
 is testing that the `sync.Once` cannot be reused after `Close()` but would it be better to mock the `tracker` to ensure that `p.tracker.recordConnMetrics()` is truly only called once?

clement0010: > Why call Close() again here, it's already closed?

We want to check if the metrics are truly measured only once. I updated the test ([cc51195](https://github.com/Twingate/kubernetes-access-gateway/pull/71/commits/cc5119575f459700ed462b7e8d5fa24e121e5d88)) to assert the `connTotal` instead - it has a simpler setup vs mocking the `tracker`

PR #79, opened 2025-07-21
feat: Add authentication via HTTP Connect method metrics

clement0010:
## Changes
- Create `InstrumentHTTPConnect` middleware that wraps `Validator` interface to measure the total authentication count and duration that happens in `ParseConnect` method


internal/connect/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+	connectTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: metrics.Namespace,
+		Name:      "tcp_connection_authentication_total",
+		Help:      "Total number of client TCP connections authenticated via HTTP Connect",
+	}, []string{"code"})
+	connectDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: metrics.Namespace,
+		Name:      "tcp_connection_authentication_duration_seconds",
+		Help:      "Duration of client TCP connections authenticated via HTTP Connect in seconds",
+		Buckets:   []float64{0.1, 0.25, 0.5, 1, 2, 5, 10, 30, 60, 120, 300, 600, 1800, 3600},
```
minhtule, line 29: These buckets are meant for long sessions using WebSocket. I think we should use typical buckets for a single HTTP request here.

internal/httpproxy/http_proxy.go
```diff
@@ -84,6 +85,7 @@ type ProxyConn struct {
 	claims *token.GATClaims
 	timer  *time.Timer
 	mu     sync.Mutex
+	start  time.Time
```
minhtule, line 88: Not sure we need a field here. We could create a timestamp at the beginning of `authenticate()` method.

internal/httpproxy/http_proxy.go
```diff
@@ -176,20 +178,22 @@ func (p *ProxyConn) authenticate() error {
...
+			httpCode = httpErr.Code
 		} else {
 			p.logger.Error("failed to parse CONNECT:", zap.Error(err))
 
-			response = httpResponseString(http.StatusBadRequest)
+			httpCode = http.StatusBadRequest
 		}
 	}
 
+	response := httpResponseString(httpCode)
```
minhtule, line 195: nit: I would move this to line 207 so the variable is declared nearer to its 1st usage.

internal/httpproxy/http_proxy.go
```diff
@@ -213,6 +217,10 @@ func (p *ProxyConn) authenticate() error {
 		return err
 	}
 
+	codeStr := strconv.Itoa(httpCode)
+	connect.RecordConnectDuration(p.start, codeStr)
+	connect.RecordConnectTotal(codeStr)
```
minhtule, line 222: I think these 3 lines should be in 1 function.

clement0010:
Closing this because we adding the same metrics in this [PR](https://github.com/Twingate/kubernetes-access-gateway/pull/71)

PR #81, opened 2025-07-28
fix: Empty `namespace` and `podname` asciiheader in `kubectl attach` session recording

clement0010:
## Changes 
- When running `kubectl attach <podname> -it`, the asciiheader for `namespace` and `podname` are empty because the `mux` handler does not match the `namespace` and `podname` pattern for `/attach` endpoint


minhtule (approved):
LGTM 👍

PR #85, opened 2025-07-30
chore(ci): Split `build` job to `build` and `test-e2e`

clement0010:
## Changes
- Update `build` job to `make build` and store the built Docker images in [workflow artifacts](https://docs.github.com/en/actions/tutorials/store-and-share-data)
 - Add `test-e2e` job which depends on build job and run e2e test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -138,13 +138,55 @@ jobs:
         VERSION_PRERELEASE: dev
       run: make build
 
-    # Run e2e tests
+    - name: Save docker images
+      run: |
+        docker images --filter=reference='twingate/kubernetes-access-gateway:*' --format '{{.Repository}}:{{.Tag}}' | \
+        while read image; do
+          filename=$(echo "$image" | sed 's|/|_|g' | sed 's|:|_|g').tar
+          docker save "$image" -o ./dist/"$filename"
+        done
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -138,13 +138,47 @@ jobs:
         VERSION_PRERELEASE: dev
       run: make build
 
-    # Run e2e tests
+    - name: Save docker images
+      run: |
+        docker images --filter=reference='twingate/kubernetes-access-gateway:*' --format '{{.Repository}}:{{.Tag}}' | \
+          xargs -I {} sh -c 'docker save {} -o ./dist/$(echo {} | tr "/:" "-").tar'
```
minhtule, line 144: This would create a tar file for each Docker image? `docker save` supports multiple images in 1 tar archive  file, which should result simpler setup.

clement0010: Yeah, we should save multiple images into 1 tar archive file. (I didn't know `docker save` could do that 🙈 ) Updated in [d6a06a3](https://github.com/Twingate/kubernetes-access-gateway/pull/85/commits/d6a06a38fcd08cbe8aadfd5f8e14c6d7412103f5)

minhtule (approved):
LGTM 👍

PR #86, opened 2025-07-30
chore: Validate asciicast in integration test

clement0010:
## Changes
- Create new assertion `AssertValidAsciicast`  to validate asciicast is valid for the integration test

### DIFF
--- a/test/integration/testutil/assertions.go
+++ b/test/integration/testutil/assertions.go
@@ -6,6 +6,7 @@ package testutil
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


test/integration/testutil/assertions.go
```diff
@@ -75,3 +82,54 @@ func AssertLogsForExec(t *testing.T, logs *observer.ObservedLogs, expectedURL, e
...
+	err := json.Unmarshal([]byte(headerLine), &header)
+	require.NoError(t, err)
+
+	assert.Equal(t, expectedHeader.Version, header.Version)
+	assert.Equal(t, expectedHeader.Width, header.Width)
+	assert.Equal(t, expectedHeader.Height, header.Height)
+	assert.Positive(t, header.Timestamp)
+	assert.Equal(t, expectedHeader.Command, header.Command)
+	assert.Equal(t, expectedHeader.User, header.User)
+	assert.Equal(t, expectedHeader.K8sMetadata, header.K8sMetadata)
```
minhtule, line 114: I would zero the `timestamp` field and then compare `assert.Equal(t, expectedHeader, header)`. If we add/remove field from the `wsproxy.AsciicastHeader` struct, we don't need to update these assertions.

minhtule (approved):
Fix & merge!

PR #89, opened 2025-07-31
feat: Support `kube-prometheus-stack` (formerly `prometheus-operator`) subchart

clement0010:
## Changes
- Add [`kube-prometheus-stack`](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#kube-prometheus-stack) (formerly `prometheus-operator`) as Helm subchart

### DIFF
--- a/deploy/gateway/values.yaml
+++ b/deploy/gateway/values.yaml
@@ -193,3 +193,18 @@ metrics:
       #     severity: warning
       #   annotations:
       #     description: Twingate Gateway authentication failure rate is {{ $value | humanizePercentage }} over the last 2 minutes.
+
+# Kube Prometheus Stack (formerly Prometheus Operator) subchart configuration
+# For more information checkout: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack
+kube-prometheus-stack:
+  enabled: false
+  
+  # Prometheus configuration
+  prometheus:
+    prometheusSpec:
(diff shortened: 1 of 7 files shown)


deploy/gateway/values.yaml
```diff
@@ -193,3 +193,18 @@ metrics:
...
+  enabled: false
+  
+  # Prometheus configuration
+  prometheus:
+    prometheusSpec:
+      # Allow Prometheus to discover all PodMonitors, ServiceMonitors, and PrometheusRules within its namespace
+      # For more information checkout: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prometheusioscrape
+      serviceMonitorSelectorNilUsesHelmValues: false
+      podMonitorSelectorNilUsesHelmValues: false
+      ruleSelectorNilUsesHelmValues: false
```
clement0010, line 209: Since the Gateway and Prometheus Operator are deployed in the same namespace, I think setting these values to `false` is easier to get started. This will set `podMonitorSelector: {}` [here](https://github.com/prometheus-community/helm-charts/blob/99bb3e471679a66be7a19bdda926096ab7700135/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L205-L209).

If we want something more complex, we could set `prometheus.prometheusSpec.podMonitorSelector.matchLabels` here anyways.

clement0010:
Closing this as we don't need to add `prometheus-operator` subchart anymore

PR #96, opened 2025-08-05
test: Add integration test for `kubectl attach`

clement0010:
## Changes
- Add integration test to check logs and asciicast in `kubectl attach`

### DIFF
--- a/test/integration/single_user_test.go
+++ b/test/integration/single_user_test.go
@@ -152,6 +152,7 @@ func TestSingleUser(t *testing.T) {
 		},
 	}
 	expectedEvents = []string{"", "hello\n"}
+
 	var exitError *exec.ExitError
 
 	output, err = user.Kubectl.CommandWithTimeout(2*time.Second, "attach", "test-pod")
--- a/test/integration/single_user_test.go
+++ b/test/integration/single_user_test.go
@@ -6,8 +6,10 @@ package integration
 import (
 	"net/http"
 	"net/url"
+	"os/exec"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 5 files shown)


test/integration/single_user_test.go
```diff
@@ -134,5 +136,34 @@ func TestSingleUser(t *testing.T) {
...
+
+	output, err = user.Kubectl.CommandWithTimeout(time.Second, "attach", "test-pod")
+	// ExitError is expected here because the command gets killed because of the timeout
+	require.ErrorAs(t, err, &exitError)
+	require.Empty(t, exitError.Stderr, "failed to execute kubectl attach")
+
+	assert.Contains(t, string(output), "hello\n")
+
+	// Wait for the logs to be flushed
+	time.Sleep(100 * time.Millisecond)
```
minhtule, line 166: Just FYI, when Go 1.25 lands later this year, we should try to use [`synctest`](https://antonz.org/go-1-25/#synthetic-time-for-testing) to avoid these sleeps.

test/integration/testutil/kubectl.go
```diff
@@ -22,15 +24,22 @@ type Kubectl struct {
...
 }
 
 // CommandWithInput is a general func to run kubectl commands with stdin input.
 func (k *Kubectl) CommandWithInput(stdinInput string, cmdOptions ...string) ([]byte, error) {
-	return k.executeKubectl(strings.NewReader(stdinInput), cmdOptions...)
+	return k.executeKubectl(context.Background(), strings.NewReader(stdinInput), cmdOptions...)
 }
 
-func (k *Kubectl) executeKubectl(stdIn io.Reader, cmdOptions ...string) ([]byte, error) {
+func (k *Kubectl) CommandWithTimeout(timeout time.Duration, cmdOptions ...string) ([]byte, error) {
```
minhtule, line 35: While this might be good enough for now, I would expose a method with a context e.g.
```go
func (k *Kubectl) CommandContext(ctx context.Context, cmdOptions ...string) ([]byte, error) { }
```
so it's more reusable (like fixing the [`contextcheck` lint issue](https://github.com/Twingate/kubernetes-access-gateway/pull/96/files#r2255661722)). I think it's common to have method with context as its first argument in Go.

minhtule (approved):
LGTM 👍

PR #122, opened 2025-08-28
chore: Remove `checksum/tlsAlternativeNames` from deployment template

clement0010:
## Changes
- After merging the TLS cert hot-reload [PR](https://github.com/Twingate/kubernetes-access-gateway/pull/111), we should remove the `checksum/tlsAlternativeNames` from `deployment.yaml` template to avoid restarting the Gateway pod when the TLS secret changes


minhtule (approved):
LGTM! 👍

PR #147, opened 2025-10-24
feat: Add PodDisruptionBudget kubernetes access gateway

twingate-blee:
> Linear Ticket URL: [DEV-3219](https://linear.app/twingate/issue/DEV-3219)
> Notion spec URL:

## Changes

This PR adds configurable PodDisruptionBudget support to the kubernetes-access-gateway Helm chart:

- Created new PDB template (`pdb.yaml`) with values-driven configuration
- Added PDB configuration to `values.yaml` (disabled by default for backward compatibility)
- Supports both `minAvailable` and `maxUnavailable` configurations
- Added comprehensive test suite (`pdb_test.yaml`) with 5 test cases covering:
  - Disabled PDB behavior
  - minAvailable configuration
  - maxUnavailable configuration
  - Metadata validation
  - Selector labels validation

## Configuration

```yaml
pdb:
  create: false        # Set to true to enable PDB
  minAvailable: 1      # Minimum number of pods that must remain available
  maxUnavailable: ""   # Maximum number/percentage of pods that can be unavailable
```

## Test Plan

- Unit tests added in `pdb_test.yaml`
- Helm chart linting passes
- Template rendering verified

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/deploy/gateway/tests/pdb_test.yaml
+++ b/deploy/gateway/tests/pdb_test.yaml
@@ -1,4 +1,4 @@
-suite: test poddisruptionbudget
+suite: Pod Disruption Budget
 templates:
   - pdb.yaml
 tests:
--- a/deploy/gateway/values.yaml
+++ b/deploy/gateway/values.yaml
@@ -92,7 +92,7 @@ extraEnvVars: []
 pdb:
   create: false
   # minAvailable: 1
-  # maxUnavailable:
+  # maxUnavailable: ""
 
 
 # TLS configuration for the Gateway


deploy/gateway/tests/pdb_test.yaml
```diff
@@ -0,0 +1,57 @@
+suite: test poddisruptionbudget
```
minhtule: nit: Not sure if Claude code picks up the convention in other files in the codebase (Windsurf does I think), but we use human words for the test suite name e.g. `Pod Disruption Budget`

minhtule (approved):
LGTM 👍

PR #190, opened 2026-02-09
feat: Support Vault AppRole authentication method

clement0010:
## Changes
- Create `newVaultAuthMethod` to return the appropriate authMethod based on the gateway config
  - Support Vault `AppRole` authentication method
- Create `newVaultClient` to return authenticated Vault client
- Add `ssh.vault.auth.approle` configuration and validation

## Notes
Support for token renewal will be added on a separate PR.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -822,26 +822,16 @@ func TestSSHCAVaultAuthConfig_Validate(t *testing.T) {
 		errContains string
 	}{
 		{
-			name:    "valid with token",
+			name:    "valid token",
 			cfg:     SSHCAVaultAuthConfig{Token: "token"},
 			wantErr: false,
 		},
 		{
-			name: "valid with appRole using secretIDFile",
+			name: "valid appRole",
 			cfg: SSHCAVaultAuthConfig{
 				AppRole: &SSHCAVaultAppRoleConfig{
 					RoleID:       "role-id",
 					SecretIDFile: "/path/to/secret-id",
-				},
-			},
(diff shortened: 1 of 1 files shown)


internal/sshhandler/vault_auth.go
```diff
@@ -0,0 +1,72 @@
...
+	return nil, nil //nolint:nilnil
+}
+
+type appRoleAuth struct {
+	roleID       string
+	secretIDFile string
+	mountPath    string
+}
+
+func (a *appRoleAuth) authenticate(ctx context.Context, client *vault.Client) (*vault.Secret, error) {
```
clement0010: We don't have a test for this because we need to write an interface wrapper for the vault client and mock it in the test...

internal/sshhandler/ca.go
```diff
@@ -106,6 +107,41 @@ func newManualCA(privateKeyFile string, logger *zap.Logger) (*caConfig, error) {
...
+	token string
+}
+
+func (a *tokenAuthMethod) Login(_ context.Context, _ *vault.Client) (*vault.Secret, error) {
+	return &vault.Secret{
+		Auth: &vault.SecretAuth{
+			ClientToken: a.token,
+			Renewable:   false,
+		},
+	}, nil
```
clement0010: I think non-root token can actually be renewed, but we need to look up the token to get its renewal metadata. I think we should support it but it is out of the scope of this PR.

minhtule (approved):
LGTM! 👍

PR #201, opened 2026-02-19
feat: Support Vault client token renewal

clement0010:
~⚠️ Branched off from #190~

## Changes
- Create a new `VaultClient` type that wraps the Vault API client with token lifecycle management that watches token expiry, re-authenticates on failure, and retries login.
- Start Vault client token renewal loop in `SSHProxy.Start()`, so we can gracefully shut down the token lifecycle loop as SSHProxy terminates.

### DIFF
--- a/internal/sshhandler/vault.go
+++ b/internal/sshhandler/vault.go
@@ -134,60 +134,60 @@ func newVault(vaultConfig *gatewayconfig.SSHCAVaultConfig, logger *zap.Logger) (
 		return nil, fmt.Errorf("failed to create vault client: %w", err)
 	}
 
-	vc := &Vault{client: client, logger: logger}
+	v := &Vault{client: client, logger: logger}
 
 	client.SetNamespace(vaultConfig.Namespace)
 
 	if vaultConfig.Auth.Token != "" {
 		client.SetToken(vaultConfig.Auth.Token)
 
-		return vc, nil
+		return v, nil
 	}
 
 	authMethod, err := newVaultAuthMethod(&vaultConfig.Auth)
 	// No auth method configured — Vault SDK falls back to VAULT_TOKEN environment variable
 	if errors.Is(err, errVaultAuthMethodNotConfigured) {
-		return vc, nil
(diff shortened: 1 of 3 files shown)


internal/sshhandler/proxy.go
```diff
@@ -294,3 +298,18 @@ func (p *SSHProxy) serveConn(ctx context.Context, conn connect.Conn) error {
 
 	return nil
 }
+
+func (p *SSHProxy) initializeVaultAuthentication(ctx context.Context) error {
+	if p.config.VaultClient == nil || p.config.VaultClient.authMethod == nil {
+		return nil
+	}
```
clement0010: When static token is used, we don't need to perform login because it's already configured on the client.

While a static token can be renewed, we are not supporting it right now because it requires a different setup. We might want support in the future, but it's out of the scope of this PR for now.

internal/sshhandler/vault.go
```diff
@@ -53,27 +65,97 @@ func newVaultClient(vaultConfig *gatewayconfig.SSHCAVaultConfig) (*vault.Client,
...
-		return client, nil
+		return vc, nil
 	}
 
 	if err != nil {
 		return nil, fmt.Errorf("failed to create Vault auth method: %w", err)
 	}
 
-	if _, err := client.Auth().Login(context.Background(), authMethod); err != nil {
-		return nil, fmt.Errorf("failed to authenticate to Vault: %w", err)
```
clement0010: Login is moved to [`sshhandler.proxy.go`](https://github.com/Twingate/kubernetes-access-gateway/pull/201/changes#diff-ba4d4a6ea7007787b2678c8a9bbff2041c7d2402f453b4236dd0d4495ae0b8e4R302-R315) so it can use the Go Context and the returned secret is also passed to the renewal loop goroutine. 

I thought of running the renewal loop here, but we need to refactor the CA code to pass context around from `Proxy.Start()`. Using `context.Background()` is not ideal either, because we can't do graceful shutdown

PR #205, opened 2026-02-25
feat: Support priorityClassName in Helm chart

### LINKED ISSUE #200: Support setting priorityClassName in Helm chart
**What is missing?**
Right now, we can not set a priorityClassName for the gateway deployment (https://github.com/Twingate/kubernetes-access-gateway/blob/master/deploy/gateway/templates/deployment.yaml).

**Why do we need it?**
This is a best practice to be able to set it to be able to have consistent behavior when resources become scarce and facing [preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/)

**Anything else we need to know?**:

clement0010:
## Changes
- Add `priorityClassName` field to the gateway deployment pod spec and to `values.yaml` alongside other pod scheduling fields
- Add Helm unit test verifying `priorityClassName` renders correctly when set

Fixes #200

### DIFF
--- a/deploy/gateway/templates/deployment.yaml
+++ b/deploy/gateway/templates/deployment.yaml
@@ -129,6 +129,6 @@ spec:
       hostAliases:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- if .Values.priorityClassName }}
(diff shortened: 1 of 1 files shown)


minhtule (approved):
LGTM 👍

PR #218, opened 2026-03-05
chore: Update `codecov` config

clement0010:
## Changes
- Update codecov `patch` coverage [`informational`](https://docs.codecov.com/docs/commit-status#informational) setting to `True` so it doesn't block PR from merging


minhtule (approved):
LGTM 👍

PR #227, opened 2026-03-11
test(integration): Add SSH integration test with Vault

clement0010:
## Changes
- Create `SetupVaultServer` to setup a Vault dev server in docker using `hashicorp/vault` image. 
  - Set up Vault SSH secrets engine with the existing test CA keys (no generated keys)
  - Set up `integration-test` Vault policy scoped to `ssh/sign/*` and `ssh/config/ca`, and enable SSH signing role
- Create `SetupVaultToken` to get a token scoped to the Vault policy instead of using the root token for authentication
- Create `SetupVaultAppRole` to enable AppRole auth and return `roleID` and `secretID` which are needed for authentication
- Update sshd test config to serve SSH host cert


test/integration/testutil/vault_server.go
```diff
@@ -0,0 +1,162 @@
...
+func SetupVaultServer(t *testing.T) (string, int) {
+	t.Helper()
+
+	// #nosec G204 -- inputs are from trusted operator configuration
+	_, err := RunCommand(exec.Command("docker", "pull", "hashicorp/vault:latest"))
+	require.NoError(t, err, "failed to pull Vault docker image")
+
+	// #nosec G204 -- inputs are from trusted operator configuration
+	output, err := RunCommand(exec.Command("docker", "run", "-d",
+		"--cap-add=IPC_LOCK",
```
minhtule: Do we really need this for testing?

minhtule (approved):
Fix & merge!

test/integration/testutil/vault_server.go
clement0010: I don't think it's needed, but the [docs example](https://hub.docker.com/r/hashicorp/vault#running-vault-in-server-mode-for-development) recommend using it:

> Note the --cap-add=IPC_LOCK: this is required in order for Vault to lock memory, which prevents it from being swapped to disk. This is highly recommended.

test/integration/testutil/vault_server.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func SetupVaultServer(t *testing.T) (string, int) {
+	t.Helper()
+
+	// #nosec G204 -- inputs are from trusted operator configuration
+	_, err := RunCommand(exec.Command("docker", "pull", "hashicorp/vault:latest"))
+	require.NoError(t, err, "failed to pull Vault docker image")
```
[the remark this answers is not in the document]
clement0010: The output of `docker run` contains the image pull logs (if image doesn't exist yet) and containerID, so we needed to pull the image first to avoid those logs and extract the containerID easily without parsing those strings.

We can use the container name instead to avoid this issue. I fixed it in [5fb0345](https://github.com/Twingate/gateway/pull/227/commits/5fb0345155b51ad9c4cfa21ff0061ca1b316396b)

test/integration/testutil/vault_server.go
minhtule: Yes, it's a security best practice in non-dev environment but I don't think we need it for testing.

PR #262, opened 2026-04-10
chore: Update README

minhtule:
## Changes

Update README to reflect renaming and that the Gateway now supports multiple protocols


aliutg (approved):
merge when ready

Issue #278, opened 2026-04-28
Support cert-manager


minhtule:
Use cert-manager to create the TLS certificate

PR #279, opened 2026-04-28
feat: Add cert-manager support for creating TLS certificate

minhtule:

## Changes

- Change `tls.autoGenerated` schema to support an object.
  + `tls.autoGenerated: true|false` still works but is deprecated.
- cert-manager integration supports self-sign CA or integrate with an existing CA

## Notes

This change is adopted from previous work in https://github.com/Twingate/gateway/pull/102


deploy/gateway/templates/tls-secret.yaml
clement0010: I think we need to check if these fields are provided. When auto-generate is disabled and these values are empty, helm install would hang.

```suggestion
tls.crt: {{ required "A valid .Values.tls.cert entry required!" .Values.tls.cert | b64enc }}
tls.key: {{ required "A valid .Values.tls.key entry required!" .Values.tls.key | b64enc }}
ca.crt: {{ required "A valid .Values.tls.ca entry required!" .Values.tls.ca | b64enc }}
```

clement0010 (approved):
Fix & merge!

PR #281, opened 2026-05-02
feat: Add `extraObjects` to Helm values file to render custom k8s manifests

minhtule:

## Changes

Add `extraObjects` to Helm values file to render custom k8s manifests

### DIFF
--- a/deploy/gateway/templates/extra-objects.yaml
+++ b/deploy/gateway/templates/extra-objects.yaml
@@ -1,0 +1,8 @@
+{{- range .Values.extraObjects }}
+---
+{{- if typeIs "string" . }}
+{{ tpl . $ }}
+{{- else }}
+{{ tpl (toYaml .) $ }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


deploy/gateway/templates/extra-objects.yaml
clement0010, line 1: I see most of the Bitnami community charts refer to this template file as `extra-list.yaml` and the value as `.Values.extraDeploys`. For example, [Kafka](https://github.com/bitnami/charts/blob/3d27e6e5048082d5e6f5b8726fed4dae9a9d00d2/bitnami/kafka/templates/extra-list.yaml) and [MySQL](https://github.com/bitnami/charts/blob/3d27e6e5048082d5e6f5b8726fed4dae9a9d00d2/bitnami/mysql/templates/extra-list.yaml). But there's also another camp like [cert-manager](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/extras-objects.yaml) and [Prometheus](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/extra-objects.yaml), which is what we currently follow. They're both essentially the same thing, but worth to note there are different conventions out there

clement0010 (approved):
LGTM 👍🏻

PR #285, opened 2026-05-11
feat: Use the address in the HTTP connect message as the upstream address

minhtule:
Issue: https://github.com/Twingate/gateway/issues/286

## Changes

- Use the address in the HTTP connect message as the upstream address
  - The user can limit the upstream's port by using Twingate Port Restriction <https://www.twingate.com/docs/resources#port-restrictions>
- Remove `upstreams` in Gateway config. Leave k8s upstream for testing for now

### DIFF
--- a/deploy/gateway/templates/gateway-configmap.yaml
+++ b/deploy/gateway/templates/gateway-configmap.yaml
@@ -26,24 +26,19 @@ data:
       privateKeyFile: "/etc/gateway/tls/tls.key"
 
     {{- $resourceEnabled := and .Values.twingate.resource .Values.twingate.resource.enabled -}}
-    {{- $k8sEnabled := and .Values.kubernetes .Values.kubernetes.enabled -}}
-    {{- $k8sUpstreams := list -}}
-    {{- if $resourceEnabled -}}
-      {{- $resourceUpstream := dict "name" (include "gateway.resourceInClusterUpstreamName" .) "inCluster" true -}}
-      {{- $k8sUpstreams = append $k8sUpstreams $resourceUpstream -}}
-    {{- end -}}
-    {{- if and $k8sEnabled .Values.kubernetes.upstreams -}}
-      {{- $k8sUpstreams = concat $k8sUpstreams .Values.kubernetes.upstreams -}}
-    {{- end }}
+    {{- $k8sEnabled := and .Values.kubernetes .Values.kubernetes.enabled }}
 
-    {{ if $k8sUpstreams -}}
+    {{ if or $resourceEnabled $k8sEnabled -}}
     kubernetes:
+      {{- if and $k8sEnabled .Values.kubernetes.upstreams }}
       upstreams:
-        {{- $k8sUpstreams | toYaml | nindent 8 }}
+        {{- .Values.kubernetes.upstreams | toYaml | nindent 8 }}
+      {{- else }} {}
+      {{- end }}
     {{ end -}}
 
     {{ with .Values.ssh -}}
-    {{ if and .enabled .upstreams -}}
+    {{ if .enabled -}}
     ssh:
       {{- with .gateway }}
       gateway:
@@ -58,14 +53,14 @@ data:
         userCertificate:
           ttl: {{ .userCertificate.ttl | quote }}
       {{- end }}
+      {{- if and .ca .ca.manual }}
 
-      {{ if and .ca .ca.manual -}}
       ca:
         manual:
           privateKeyFile: "/etc/gateway/ssh/ca-private-key"
-      {{ end -}}
+      {{- end }}
+      {{- if and .ca .ca.vault }}
 
-      {{ if and .ca .ca.vault -}}
       ca:
         vault:
           address: {{ .ca.vault.address | quote }}
@@ -105,9 +100,6 @@ data:
             mount: {{ . | quote }}
             {{- end }}
           {{- end }}
-      {{ end -}}
-
-      upstreams:
-        {{- .upstreams | toYaml | nindent 8 }}
+      {{- end }}
     {{ end -}}
     {{ end -}}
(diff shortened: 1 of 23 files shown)


internal/config/config.go
```diff
@@ -285,12 +260,8 @@ func (k *KubernetesUpstream) Validate() error {
...
 
-	if !k.InCluster && k.Address == "" {
-		return fmt.Errorf("%w: address is required when inCluster is false", ErrRequired)
-	}
-
-	if !k.InCluster && k.BearerToken == "" && k.BearerTokenFile == "" {
-		return fmt.Errorf("%w: either bearerToken or bearerTokenFile is required when inCluster is false", ErrRequired)
+	if k.BearerToken == "" && k.BearerTokenFile == "" {
+		return fmt.Errorf("%w: either bearerToken or bearerTokenFile is required", ErrRequired)
 	}
```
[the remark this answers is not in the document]
minhtule: we didn't validate this previously. This code would be eventually removed.

deploy/gateway/values.schema.json
```diff
@@ -94,18 +85,10 @@
...
-                {"required": ["bearerTokenFile"]}
-              ]
-            },
+            "oneOf": [
+              {"required": ["bearerToken"]},
+              {"required": ["bearerTokenFile"]}
+            ],
             "additionalProperties": false
           }
         }
```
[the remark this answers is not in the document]
minhtule: Fixed in https://github.com/Twingate/gateway/pull/285/commits/63a17ae224c9e5e4a67ab65e586e20c0bdb65bb5

internal/httphandler/http_proxy.go
```diff
@@ -44,7 +44,7 @@ func NewProxy(cfg Config) (*Proxy, error) {
 	logger := cfg.logger
 
 	// create TLS configuration for upstream
-	caCert, err := os.ReadFile(cfg.upstream.CAFile)
+	caCert, err := os.ReadFile(cfg.caFile)
 	if err != nil {
 		return nil, err
```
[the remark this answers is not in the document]
minhtule: this is same behavior as before.

internal/config/config.go
clement0010: Let's update the comment here?

clement0010: Let's update the comment here?

internal/config/config.go
```diff
@@ -164,12 +159,6 @@ type SSHCAVaultAWSConfig struct {
 	Nonce         string `yaml:"nonce,omitempty"`
 }
 
-type SSHUpstream struct {
```
clement0010, line 167: We should also remove the usage in https://github.com/Twingate/gateway/blob/6bb05ed7a5fb138025b3f185634233113d3172d5/tools/local/main.go#L221-L223

deploy/gateway/templates/gateway-configmap.yaml
```diff
@@ -26,24 +26,19 @@ data:
...
-    {{ if $k8sUpstreams -}}
+    {{ if or $resourceEnabled $k8sEnabled -}}
     kubernetes:
+      {{- if and $k8sEnabled .Values.kubernetes.upstreams }}
       upstreams:
-        {{- $k8sUpstreams | toYaml | nindent 8 }}
+        {{- .Values.kubernetes.upstreams | toYaml | nindent 8 }}
+      {{- else }} {}
+      {{- end }}
     {{ end -}}
```
clement0010: Would it be nicer to do
```yaml
{{ if and $k8sEnabled .Values.kubernetes.upstreams -}}
kubernetes:
  upstreams:
    {{- .Values.kubernetes.upstreams | toYaml | nindent 8 }}
{{ else if or $resourceEnabled $k8sEnabled -}}
kubernetes: {}
{{ end -}}
```
?

internal/httphandler/config.go
```diff
@@ -12,43 +12,53 @@ import (
...
 	}
 
-	inClusterConfig, err := rest.InClusterConfig()
+	upstream := k8sConfig.Upstreams[0]
+	cfg.bearerToken = upstream.BearerToken
+	cfg.bearerTokenFile = upstream.BearerTokenFile
+	cfg.caFile = upstream.CAFile
+
+	if upstream.Address != "" {
+		logger.Warn("kubernetes.upstreams[].address is ignored; upstream routing comes from the GAT",
```
clement0010: Question: Is there a reason why we still want to keep the `upstream.Address` field if it's not being used?

clement0010 (commented):
I think we should also include a GitHub Issue in the PR description for tracking purposes right?

internal/httphandler/config.go
minhtule: No, we can get rid of it!

clement0010 (approved):
LGTM 👍🏻

internal/httphandler/config.go
```diff
@@ -12,43 +12,46 @@ import (
...
+		}
 
-func GetInClusterConfig(upstream *config.KubernetesUpstream) (*config.KubernetesUpstream, error) {
-	if !upstream.InCluster {
-		return upstream, nil
+		return cfg, nil
 	}
 
-	inClusterConfig, err := rest.InClusterConfig()
+	upstream := k8sConfig.Upstreams[0]
```
sghiocel, line 38: nit: the comment from previous code `// Multiple upstreams support will be added soon!` was removed (accidental?) - maybe you intended to keep it

sghiocel (approved):

internal/httphandler/config.go
minhtule: That comment is meant to explain why we only use the first upstream with `k8sConfig.Upstreams[0]`. Now that there is no `upstreams` in the config, I think it's not needed. Plus, we'll support multiple upstreams for k8s at some point but probably not "soon" because there is no use case/user request yet 😬

PR #291, opened 2026-05-14
feat: Simplify cert-manager integration

minhtule:

## Changes

cert-manager integration expects the user to create the Issuer as part of the Gateway Helm chart's `extraObjects` (or a different Helm chart). This removes (1) self-sign CA mode with cert-manager and (2) intermediate CA from the chart.

## Notes

Breaking change for the cert-manager integration introduced in v0.15.0 (#279). Helm self-signing remains the default and is backward compatible for users who relied on defaults or only set `tls.autoGenerated.enabled`, but values files that explicitly carry the v0.15.0 `engine:` or `certManager.*` keys will fail schema validation and need to be updated.


deploy/gateway/values.yaml
```diff
@@ -191,18 +191,19 @@ tls:
...
-    # Mechanism to generate the certificates (allowed values: helm, cert-manager)
-    engine: helm
-    # `certManager` section is only used when `engine` is set to `cert-manager`.
     certManager:
-      existingIssuer: ""
-      existingIssuerKind: Issuer
-      keySize: 2048
-      keyAlgorithm: RSA
-      duration: 2160h
-      renewBefore: 360h
```
minhtule, line 205: We can use cert-manager `Certificate` [default values](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.Certificate).

deploy/gateway/templates/_helpers.tpl
```diff
@@ -80,15 +80,18 @@ Create the name of the TLS secret to use
...
+otherwise helm.
 */}}
 {{- define "gateway.tlsEngine" -}}
 {{- if kindIs "bool" .Values.tls.autoGenerated -}}
 {{- if .Values.tls.autoGenerated -}}helm{{- end -}}
 {{- else if .Values.tls.autoGenerated -}}
 {{- if .Values.tls.autoGenerated.enabled -}}
-{{- default "helm" .Values.tls.autoGenerated.engine -}}
+{{- $cm := .Values.tls.autoGenerated.certManager -}}
+{{- if and $cm $cm.issuerRef $cm.issuerRef.name -}}cert-manager{{- else -}}helm{{- end -}}
```
ThatsMrTalbot: It may be better to keep this as an explicit toggle, like `tls.autoGenerated.certManager.enabled`, you could then use the [required](https://helm.sh/docs/chart_template_guide/function_list#required) function to have the chart error if cert-manager is enabled without providing an issuerRef kind/name.

Personally, I don't like when helm charts "infer" my intention instead of explicit controls and validation.

deploy/gateway/templates/cert-manager.yaml
```diff
@@ -10,15 +10,7 @@ metadata:
     {{- include "gateway.labels" . | nindent 4 }}
 spec:
   secretName: {{ include "gateway.tlsSecretName" . }}
-  duration: {{ $cm.duration }}
-  renewBefore: {{ $cm.renewBefore }}
   commonName: {{ $fullname }}
```
ThatsMrTalbot: Some issuers, like Vault, validate the common name. For example by default Vault will validate the value is both a valid hostname AND passes validation for allowed hostnames.

Having the capability to override it to a specified value would be useful.

deploy/gateway/templates/cert.yaml
```diff
@@ -10,15 +10,7 @@ metadata:
     {{- include "gateway.labels" . | nindent 4 }}
 spec:
   secretName: {{ include "gateway.tlsSecretName" . }}
-  duration: {{ $cm.duration }}
-  renewBefore: {{ $cm.renewBefore }}
```
ThatsMrTalbot, line 14: I think these dials are useful and should be kept, maybe in an if statement around each one so they only appear if you set them

ThatsMrTalbot (commented):
Hiya! I have some feedback on this PR based on our usage of the gateway

ThatsMrTalbot (approved):
This looks like it should cover our use cases, thanks for looking at my feedback!

deploy/gateway/templates/_helpers.tpl
clement0010:
```suggestion
Return "true" when the chart should render cert-manager Certificate resources.
```

deploy/gateway/templates/cert-manager.yaml
```diff
@@ -10,15 +10,11 @@ metadata:
...
-  commonName: {{ $fullname }}
-  subject:
-    organizations:
-      - Twingate
-  privateKey:
-    algorithm: {{ $cm.keyAlgorithm }}
-    size: {{ $cm.keySize }}
+  issuerRef:
+    name: {{ required "tls.autoGenerated.certManager.issuerRef.name is required when tls.autoGenerated.certManager.enabled is true" $cm.issuerRef.name }}
+    kind: {{ $cm.issuerRef.kind }}
```
clement0010, line 15: Since we are already doing the same thing for `group`, I think we should fallback to a default value: `Issuer` when it's an empty string?

clement0010 (approved):
LGTM 👍🏻

ekampf (approved):

deploy/gateway/templates/cert-manager.yaml
minhtule: That's a good callout. It was my mistake to add a default for the `group` field here, since `values.yaml` already specifies one. Duplicating it in the template risks inconsistency. We also don't use `default` in other templates when `values.yaml` already specifies a value. I'll remove the default for `group` field.

PR #293, opened 2026-05-18
feat: Resolve sharded URL at startup

clement0010:
Issue: https://github.com/Twingate/gateway/issues/292

## Changes

- Resolve sharded Twingate host at config load time by sending a `HEAD` request to the JWKS endpoint and extract the `Location` header
  - `resolveTwingateHostname` extract the `Location` header and don't follow the redirect
  - `stripNetworkPrefix` strips the network prefix from the resolved hostname (e.g., `acme.us1.twingate.com` → `us1.twingate.com`)
  - Falls back to the original host on any connection error
- Support sharded hosts in JWT issuer validation via suffix matching in `getIssuer()`

Note: Sharded host is not supported in production yet. This is preparation work, so the operator is ready when sharding is enabled in production.


internal/config/config.go
```diff
@@ -187,6 +189,44 @@ func Load(path string) (*Config, error) {
...
+	if err != nil {
+		return defaultHost
+	}
+
+	return location.Hostname()
+}
+
+func (c *Config) ResolveTwingateHost() {
+	targetURL := fmt.Sprintf("https://%s.%s/api/v1/jwk/ec", c.Twingate.Network, c.Twingate.Host)
+	resolvedHostname := resolveTwingateHostname(targetURL, c.Twingate.Host, 3)
```
clement0010: With this configuration we get:

| Step      | What           | Duration |
|-----------|----------------|----------|
| Attempt 0 | HEAD → timeout | 1s       |
| Backoff 0 | 2^0 × 1s       | 1s       |
| Attempt 1 | HEAD → timeout | 1s       |
| Backoff 1 | 2^1 × 1s       | 2s       |
| Attempt 2 | HEAD → timeout | 1s       |
| Backoff 2 | 2^2 × 1s       | 4s       |
| Attempt 3 | HEAD → timeout | 1s       |

**Total = 11s**

I think 11s is fine? Alternatively, we can shorten the backoff to 0.5s or set the max attempt to 2, to squeeze within the 10s window

internal/token/parser.go
```diff
@@ -22,6 +23,16 @@ var allowedIssuerByHost = map[string]string{
 	"twingate.com":  "twingate",
 }
 
+func getIssuer(host string) string {
+	for globalHost, issuer := range allowedIssuerByHost {
```
minhtule: `globalHost` doesn't sound right. I would rename the map to somthing like
```go
for baseDomain, issuer := range allowedIssuerByDomain { 
```

internal/config/config.go
minhtule: - Yes, I would lower the max retries to 2. Maybe also increase the timeout to 2s.
- Why is the number of retries passed in as an argument but the 1s timeout is hard coded [here](https://github.com/Twingate/gateway/pull/293/changes#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR198)?

internal/config/config_test.go
```diff
@@ -13,6 +15,92 @@ import (
...
+			t.Fatal("should not follow redirect to shard server")
+		default:
+		}
+	})
+
+	t.Run("returns default host on connection error", func(t *testing.T) {
+		result := resolveTwingateHostname("http://127.0.0.1:1/api/v1/jwk/ec", "twingate.com", 0)
+
+		assert.Equal(t, "twingate.com", result)
+	})
```
minhtule, line 102: How long does this test take? does it need to wait for 1s for the request to time out?

minhtule (approved):
Fix & merge!

internal/config/config.go
clement0010: > Why is the number of retries passed in as an argument but the 1s timeout is hard coded [here](https://github.com/Twingate/gateway/pull/293/changes#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR198)?

We are setting the same value (1 second) for the test as well, so Go lint complains: `internal/config/config.go:196:61: resolveTwingateHostname - timeout always receives 1 * time.Second (1000000000) (unparam)`, whereas `retryMax` is set to 0 for all tests

internal/config/config_test.go
clement0010: > How long does this test take? does it need to wait for 1s for the request to time out?

It returns immediately. The DNS should resolve immediately instead of waiting for timeout.

internal/config/config.go
```diff
@@ -187,6 +189,44 @@ func Load(path string) (*Config, error) {
 	return cfg, nil
 }
 
+func stripNetworkPrefix(hostname, network string) string {
+	return strings.TrimPrefix(hostname, network+".")
+}
+
+func resolveTwingateHostname(targetURL, defaultHost string, retryMax int) string {
```
sghiocel: Minor: It would be nice to have some logging to be able to see the behavior/result of this function, especially since it can potentially delay startup by a few seconds.

sghiocel (approved):
Added one minor comment, decide if you want to address it now, otherwise looks good to merge.

PR #296, opened 2026-05-21
refactor: HTTP proxy to support multiple resource type

clement0010:

- Issue: https://github.com/Twingate/gateway/issues/297

## Changes

Refactors the HTTP proxy layer to separate protocol-agnostic concerns from Kubernetes-specific logic, enabling future support for multiple resource types (e.g. generic HTTP proxying alongside Kubernetes).

- **Extract `internal/httpproxy/` package**: Moves the generic HTTP server, connection context, and audit middleware out of `internal/httphandler/` into a new `internal/httpproxy/` package. The `httpproxy.Proxy` owns the `http.Server`, `ConnContext` wiring, metrics middleware, and audit middleware — all resource-type-agnostic.
- **Introduce `KubernetesHandler`**: Replaces the former monolithic `httphandler.Proxy` with a focused `KubernetesHandler` that implements `http.Handler`. It owns only Kubernetes-specific concerns: upstream TLS transport, bearer token auth, and K8s impersonation header rewriting.
- **Decouple metrics registration from instrumentation**: Splits `metrics.RoundTripper()` into `RegisterRoundTripperMetrics()` (register once) and `InstrumentRoundTripper()` (wrap per-transport). This allows multiple resource-type handlers to share a single set of Prometheus collectors without double-registration panics.
- **Simplify audit middleware**: Converts the callback-based `handlerWithAuditLogger` pattern to standard `http.Handler` composition. The audit logger is now passed via context (`AuditLoggerFromContext`), removing the custom function signature coupling.
- **Improve test isolation**: Tests for `KubernetesHandler` use `httptest` directly instead of `net.Pipe` + channel-based plumbing. Audit middleware tests use plain `http.Handler` instead of mock objects with custom signatures.
- Rename `httphandler` to `kuberneteshandler`


internal/httphandler/http_proxy.go
```diff
@@ -5,110 +5,56 @@ package httphandler
...
-				r.Out.Header.Add("Impersonate-Group", group)
-			}
-		},
-		Transport: metrics.RoundTripper(metrics.RoundTripperConfig{
-			Registry: cfg.registry,
-			Next:     transport,
-		}),
+				handler.Rewrite(r, conn)
+			},
+			Transport: metrics.InstrumentRoundTripper(collectors, transport),
```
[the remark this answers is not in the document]
clement0010: This issue was fixed in `Go 1.22`: https://go.dev/blog/loopvar-preview

internal/httpproxy/audit_middleware.go
```diff
@@ -72,10 +73,18 @@ func (rw *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 var _ http.Flusher = &responseWriter{}  // Support HTTP streaming
 var _ http.Hijacker = &responseWriter{} // Support WebSocket streaming
 
-type handlerWithAuditLogger func(w http.ResponseWriter, r *http.Request, conn *connect.ProxyConn, auditLogger *zap.Logger)
```
clement0010, line 75: The reason we had this type was that we needed to expose the audit logger in the K8s HTTP handler for session recording, but other HTTP handlers (e.g. WebApp) might not need it. 

I think it's nicer for the audit logger to be passed in the request context instead, so each handler retrieves the audit logger from context via `AuditLoggerFromContext` only if it needs it.

internal/httphandler/http_proxy_test.go
clement0010, line 1: This test is moved to `internal/httphandler/kubernetes_test.go`

internal/metrics/round_tripper.go
```diff
@@ -10,43 +10,50 @@ import (
...
 )
 
-type RoundTripperConfig struct {
-	Registry *prometheus.Registry
-	Next     http.RoundTripper
+type RoundTripperCollectors struct {
+	requestsTotal   *prometheus.CounterVec
+	activeRequests  *prometheus.GaugeVec
+	requestDuration *prometheus.HistogramVec
 }
```
clement0010: I refactored these metric collectors so that other roundtrippers across different resource types can reuse them later

internal/kuberneteshandler/kubernetes_test.go
```diff
@@ -0,0 +1,233 @@
...
+	handler, err := NewKubernetesHandler(Config{
+		bearerToken:            "mock-token",
+		caFile:                 "../../test/data/api_server/tls.crt",
+		auditLog:               &config.AuditLogConfig{},
+		roundTripperCollectors: metrics.RegisterRoundTripperMetrics(prometheus.NewRegistry()),
+		logger:                 zap.NewNop(),
+	})
+	require.NoError(t, err)
+
+	// Create a ProxyConn in context so both Rewrite (connect.Conn) and ServeHTTP (*connect.ProxyConn) work
```
minhtule: nit: Rewrite also casts to `connect.ProxyConn`. I think we can drop this comment.

internal/kuberneteshandler/kubernetes.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			conn, ok := r.In.Context().Value(httpproxy.ConnContextKey).(*connect.ProxyConn)
+			if !ok {
+				cfg.logger.Error("Failed to retrieve net.Conn from context")
+
+				return
+			}
```
minhtule: This is extracted from the existing code but I think we should simplify it by removing the checking. If the type assertion fails, it's a programmatic bug and the request should panic instead of still being forwarded to the upstream. If the linter complains, we should silent it.

internal/kuberneteshandler/kubernetes.go
clement0010, line 1: Most of the logic in this file is copied from `internal/httphandler/http_proxy.go`

internal/kuberneteshandler/kubernetes.go
```diff
@@ -0,0 +1,156 @@
...
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	auditLogger := httpproxy.AuditLoggerFromContext(r.Context())
+
+	conn, ok := r.Context().Value(httpproxy.ConnContextKey).(*connect.ProxyConn)
+	if !ok {
+		auditLogger.Error("Failed to retrieve proxy connection from context")
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+
+		return
+	}
```
clement0010: This code is new

internal/kuberneteshandler/kubernetes_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+	"gateway/internal/config"
+	"gateway/internal/connect"
+	"gateway/internal/httpproxy"
+	"gateway/internal/metrics"
+	"gateway/internal/token"
+	"gateway/test/data"
+)
+
+func TestHandler_ServeHTTP(t *testing.T) {
```
clement0010, line 29: Only this test is copied over from `internal/httphandler/http_proxy_test.go`.  The rest are new tests

internal/httpproxy/audit_middleware.go
```diff
@@ -72,10 +73,18 @@ func (rw *responseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
...
+type auditLoggerKey struct{}
+
+func AuditLoggerFromContext(ctx context.Context) *zap.Logger {
+	if logger, ok := ctx.Value(auditLoggerKey{}).(*zap.Logger); ok {
+		return logger
+	}
+
+	return zap.NewNop()
+}
```
minhtule, line 83: Instead of returning no-op logger (no audit log), I think it's better to panic here if the caller is not using the `httpproxy` but want audit log. It's a programmatic error and shouldn't silently continue.

internal/httpproxy/proxy.go
```diff
@@ -0,0 +1,68 @@
...
+	mux.Handle("/", handler)
+
+	httpServer := &http.Server{
+		// G112 - Protect against Slowloris attack
+		ReadHeaderTimeout: 5 * time.Second,
+		Handler:           mux,
+		// add the net.Conn to the context so we can track this connection, this context
+		// will be merged with and retrievable in the http.Request that is passed in to the Handler func and
+		// since our custom listener provided a wrapped net.Conn (ProxyConn), its fields will be
+		// available, specifically the identity information parsed from CONNECT
```
minhtule: nit: I think we should drop this comment because it mostly describes how `http.Server` uses this context.

internal/httpproxy/proxy.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+
+	"gateway/internal/metrics"
+)
+
+type connContextKey string
+
+const ConnContextKey connContextKey = "CONN_CONTEXT"
```
minhtule: Instead of exposing the context key, we could provide a getter function like the `AuditLoggerFromContext()`

internal/kuberneteshandler/kubernetes_test.go
minhtule: This test is more than a unit test and almost like an integration test. It tests the wiring of the `http.Server` and `httputil.ReverseProxy` which is better covered by integration tests in `test/integration/kubernetes_test.go` (running an actual k8s API server). Now that the code is more modular and easier to unit test, I think we should remove this test. Let's also see what @sghiocel thinks!

internal/metrics/round_tripper.go
```diff
@@ -10,43 +10,50 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-type RoundTripperConfig struct {
-	Registry *prometheus.Registry
-	Next     http.RoundTripper
+type RoundTripperCollectors struct {
```
minhtule: I would name this struct `RoundTripperMetrics` so it fits nicely with the `RegisterRoundTripperMetrics()` function.
```go
type RoundTripperMetrics struct { ... }
func RegisterRoundTripperMetrics(registry) *RoundTripperMetrics
```
"Metric" also sounds less scary/technical than "collector" 🙈

minhtule (approved):
Fix & merge!

internal/metrics/round_tripper.go
clement0010: Updated in [0ab84a0](https://github.com/Twingate/gateway/pull/296/commits/0ab84a0ef6afe598d02d7167cf981a4f3cf9bcbd)

internal/httpproxy/audit_middleware.go
clement0010: Updated in [d636743](https://github.com/Twingate/gateway/pull/296/commits/d63674332330790659c34929f755f69a9a95f36d)

internal/httpproxy/proxy.go
clement0010: Updated in [d636743](https://github.com/Twingate/gateway/pull/296/commits/d63674332330790659c34929f755f69a9a95f36d)

internal/kuberneteshandler/kubernetes.go
clement0010: Updated to use `ProxyConnFromContext()` in [d636743](https://github.com/Twingate/gateway/pull/296/commits/d63674332330790659c34929f755f69a9a95f36d), which would panic if the type assertion fails

sghiocel (approved):
Looks good to me!

internal/kuberneteshandler/kubernetes_test.go
sghiocel: I agree, we can remove this test

PR #310, opened 2026-05-28
feat: Support HTTP web app

clement0010:

- Issue: #309

## Changes

- Add new `webapphandler` package to handle HTTP web application
  - Add web app configuration and validation
  - Add template utility to parse and evaluate `prefix {{twingate.<key>}} suffix` template
- Supporting running one HTTP proxy per resource type and update `connect` package's channel-based routing by `ResourceType` instead of `TransportProtocol` type
- Add `resourceType` label to roundtripper and http middleware metrics

## Notes


internal/webapphandler/web_app.go
```diff
@@ -0,0 +1,76 @@
...
+func NewHandler(cfg Config) *Handler {
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			conn := httpproxy.ProxyConnFromContext(r.In.Context())
+
+			if err := rewrite(r, conn, cfg.headers); err != nil {
+				cfg.logger.Error("failed to rewrite headers", zap.Error(err))
+			}
+		},
+		Transport: metrics.InstrumentRoundTripper(cfg.roundTripperMetrics, "webapp", http.DefaultTransport),
```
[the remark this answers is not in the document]
clement0010: If we don't have metrics here, `httputil.ReverseProxy` will also use the `http.DefaultTransport`

internal/proxy/proxy.go
```diff
@@ -45,9 +47,14 @@ type Proxy struct {
...
 	var httpProxy *httpproxy.Proxy
 
-	if config.Kubernetes != nil {
-		roundTripperMetrics := metrics.RegisterRoundTripperMetrics(registry)
+	handlers := make(map[string]http.Handler)
+
+	var roundTripperMetrics *metrics.RoundTripperMetrics
+	if config.Kubernetes != nil || config.WebApp != nil {
+		roundTripperMetrics = metrics.RegisterRoundTripperMetrics(registry)
+	}
```
clement0010, line 58: When we have 3+ HTTP resource types, I think we should refactor the config to register the handler + metrics more nicely and avoid this check... It's out of the scope of this PR for now

internal/httpproxy/template/template.go
clement0010, line 1: I didn't use Go's `text/template` package because it doesn't support the `{{namespace.key}}` syntax — it requires a leading dot (e.g., `{{ .twingate.key}}`).

internal/httpproxy/template/template.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+	"unicode"
+)
+
+const (
+	allowedNamespace = "twingate"
```
clement0010, line 15: We can extend this to support more namespaces like `internal` and `external` in the future, but I don't think we need to do it for now 🤔

internal/httpproxy/utils/parser/parser.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var templateRe = regexp.MustCompile(
+	`^([^{}]*)` + // prefix (no brackets allowed)
+		`{{\s*` + // opening braces
+		`([a-zA-Z0-9_-]+)` + // namespace
+		`\.` +
+		`([a-zA-Z0-9_-]+)` + // key
+		`\s*}}` + // closing braces
+		`([^{}]*)$`, // suffix (no brackets allowed)
```
clement0010: FYI, This won't allow header templates with multiple variables like `X-Header: {{twingate.username}} {{twingate.groups}}`. Not sure there's a real use case for it 🤔

We could support it with some additional changes, but I'd rather not add complexity to the parser unless we need to. We can follow-up PR if a use case comes up.

internal/metrics/round_tripper.go
```diff
@@ -22,55 +34,59 @@ func RegisterRoundTripperMetrics(registry *prometheus.Registry) *RoundTripperMet
 			Namespace: Namespace,
 			Name:      "api_server_requests_total",
 			Help:      "Total number of requests from Gateway to API Server processed",
-		}, []string{"type", "method", "code"}),
+		}, []string{labelResourceType, "type", "method", "code"}),
```
clement0010, line 38: We also need to update the [`grafana-dashboard.json`](https://github.com/Twingate/gateway/blob/b817326d82e762eaa28ee43eea405b7000035055/deploy/gateway/dashboards/gateway-dashboard.json) file to support this new label, i'll follow up with a separate PR instead

internal/connect/conn.go
```diff
@@ -105,6 +108,10 @@ func (p *ProxyConn) TransportProtocol() TransportProtocol {
 	return TransportTLS
 }
 
+func (p *ProxyConn) ShouldUpgradeTLS() bool {
+	return p.TransportProtocol() == TransportTLS && p.GATClaims().Resource.Type != token.ResourceTypeWebApp
+}
```
minhtule: This is okayish for now but I think eventually this logic should be on the GAT's resource. It knows whether the inner connection should be not encrypted or TLS/SSH.

internal/webapphandler/handler.go
minhtule, line 1: There is no need to prefix files under `webapphandler` package with `web_app`. There should be `handler.go` and `config.go`

internal/webapphandler/handler.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func NewHandler(cfg Config) *Handler {
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(r *httputil.ProxyRequest) {
+			conn := httpproxy.ProxyConnFromContext(r.In.Context())
+
+			if err := rewrite(r, conn, cfg.headers); err != nil {
+				cfg.logger.Error("failed to rewrite headers", zap.Error(err))
+			}
```
minhtule, line 33: I think when header rewriting fails, it's safer to fail the request.

internal/webapphandler/web_app.go
```diff
@@ -0,0 +1,76 @@
...
+	geoLoc := ""
+	if claims.Device.Location != (token.GeoIPLocation{}) {
+		geoLoc = fmt.Sprintf("%v,%v", claims.Device.Location.Lat, claims.Device.Location.Lon)
+	}
+
+	variables := map[string]string{
+		"jwt":          conn.GetToken(),
+		"username":     claims.User.Username,
+		"groups":       strings.Join(claims.User.Groups, ","),
+		"clientGeoLoc": geoLoc,
```
minhtule: Let's call this `clientGeoLatLong` for clarity.

internal/config/config.go
```diff
@@ -436,6 +451,26 @@ func (v *SSHCAVaultConfig) Validate() error {
 	return nil
 }
 
+func (w *WebAppConfig) Validate() error {
+	allowedWebAppKeys := []string{
+		"jwt", "username", "groups", "clientGeoLoc",
```
minhtule: - These should be constants that we can use in the `webapphandler/web_app.go` file.
- This should be a global constant slice.

internal/metrics/round_tripper.go
```diff
@@ -22,55 +34,59 @@ func RegisterRoundTripperMetrics(registry *prometheus.Registry) *RoundTripperMet
...
 			Name:      "api_server_active_requests",
 			Help:      "Number of currently active requests from Gateway to API Server",
-		}, []string{"type"}),
+		}, []string{labelResourceType, "type"}),
 
 		requestDuration: prometheus.NewHistogramVec(
 			prometheus.HistogramOpts{
 				Namespace: Namespace,
 				Name:      "api_server_request_duration_seconds",
 				Help:      "Measures the initial HTTP request-response latency between Gateway and API Server in seconds. For HTTP streaming, WebSocket, and SPDY connections, this metric captures only the setup time and not the duration of the data transfer.",
```
minhtule: Probably should update the help text to something more generic like "HTTP server" instead of "API server"

internal/proxy/proxy.go
minhtule: Would it be simpler if each resource type has its own HTTP server + proxy handler instead of all them sharing the same HTTP server? In that case, I think the HTTP proxy could still stay generic and doesn't have to know about the resource type.

internal/httpproxy/utils/parser/parser.go
minhtule: nit: "bracket" and "brace" are similar. We should just stick with "braces".

internal/httpproxy/parser/parser.go
```diff
@@ -0,0 +1,81 @@
...
+		`([^{}]*)$`, // suffix (no brackets allowed)
+)
+
+type Template struct {
+	prefix string
+	key    string
+	suffix string
+}
+
+// NewTemplate parses a string like "<prefix> {{<namespace>.<key>}} <suffix>" into a Template.
```
minhtule: nit: I would add that if there is no template variable (just a static string). The `key` and `suffix` are empty and the `prefix` is the static string.

internal/httpproxy/template/template.go
minhtule: I think this parser should be in `webapphandler` than here. Also it doesn't need to be under `/utils/parser` path. I would put it in `internal/webapphandler/template.go`

internal/httpproxy/template/template.go
```diff
@@ -0,0 +1,81 @@
...
+	prefix, namespace, key, suffix := match[1], match[2], match[3], match[4]
+
+	if namespace != allowedNamespace {
+		return nil, fmt.Errorf("%w: unsupported namespace %q", ErrInvalidTemplate, namespace)
+	}
+
+	return &Template{
+		prefix: strings.TrimLeftFunc(prefix, unicode.IsSpace),
+		key:    key,
+		suffix: strings.TrimRightFunc(suffix, unicode.IsSpace),
```
minhtule, line 82: If we trim spaces from a template string, should we trim static string ([line 48](https://github.com/Twingate/gateway/pull/310/changes#diff-f55e4aa4e77724cf27a57a8c88f1f40f5bc7ffa7590c2718318bc24ddc43477dR48)) too?

internal/httpproxy/template/template.go
clement0010: My thinking is that this parser could also be shared with other protocols (e.g. MCP) that support header rewrite or other applications that need to parse a template. 

We can move this under `webapphandler` for now, but I think we might still need to make this a shared utility package eventually 🤔

internal/httpproxy/template/template.go
clement0010: Yes, we should! Fixed in [0c37a75](https://github.com/Twingate/gateway/pull/310/commits/0c37a75f5ab1b7a67a4b985fa0582a16cd2b04dd)

internal/config/config.go
clement0010: Updated in [4016e4a](https://github.com/Twingate/gateway/pull/310/commits/4016e4aff19e021f31c601fd2e6edef69f46bddd)

internal/webapphandler/handler.go
clement0010: Updated in [0699bd5](https://github.com/Twingate/gateway/pull/310/commits/0699bd5c513675f0233a91863d01fe45bf0a8960)

internal/proxy/proxy.go
clement0010: I updated the code to support each resource type to own its HTTP server. [e7ceee9](https://github.com/Twingate/gateway/pull/310/commits/e7ceee93702e366c0a95e43049d7f529a61d5c65). This should simplify the setup quite a bit

internal/connect/conn.go
clement0010: After updating the code to HTTP proxy per resource type, I think it's simpler to have this check on the GAT claims. Also, we could remove the `TransportProtocol` type to simplify the check.

internal/webapphandler/handler.go
```diff
@@ -0,0 +1,82 @@
...
+		Host:   conn.GetAddress(),
+	}
+	r.SetURL(targetURL)
+
+	claims := conn.GATClaims()
+
+	clientLocation := claims.Device.Location
+
+	clientGeoLatLong := ""
+	if clientLocation != (token.GeoIPLocation{}) {
```
minhtule:
```go
clientLocation.Lat != 0 || clientLocation.Lon != 0
```
check is better because this would emit `0,0` when the `GeoIPLocation`'s other fields like country are non-null.

internal/metrics/round_tripper.go
```diff
@@ -4,12 +4,24 @@
...
+// Metric label names.
+
+const labelResourceType = "resource_type"
+
+// Resource type values.
+
+const (
+	ResourceTypeKubernetes = "kubernetes"
+	ResourceTypeWebApp     = "web_app"
+)
```
minhtule, line 24: nit: we could create a `ResourceType` for these enum values.

internal/webapphandler/handler.go
```diff
@@ -0,0 +1,82 @@
...
+
+	variables := map[string]string{
+		template.JWT:              conn.GetToken(),
+		template.Username:         claims.User.Username,
+		template.Groups:           strings.Join(claims.User.Groups, ","),
+		template.ClientGeoLatLong: clientGeoLatLong,
+		template.ClientGeoCity:    clientLocation.City,
+		template.ClientGeoRegion:  clientLocation.Region,
+		template.ClientGeoCountry: clientLocation.Country,
+	}
```
minhtule, line 62: When we add a new template key, we need to update this mapping. Otherwise, all the requests would fail. We could enforce this via unit test e.g.
```go
func buildVariables(conn *connect.ProxyConn) map[string]string {
	claims := conn.GATClaims()
	loc := claims.Device.Location

	latLong := ""
	if loc.Lat != 0 || loc.Lon != 0 { // the #2 fix, folded in
		latLong = fmt.Sprintf("%v,%v", loc.Lat, loc.Lon)
	}

	return map[string]string{
		template.JWT:              conn.GetToken(),
		template.Username:         claims.User.Username,
		template.Groups:           strings.Join(claims.User.Groups, ","),
		template.ClientGeoLatLong: latLong,
		template.ClientGeoCity:    loc.City,
		template.ClientGeoRegion:  loc.Region,
		template.ClientGeoCountry: loc.Country,
	}
}

func TestBuildVariables_CoversAllowedKeys(t *testing.T) {
	conn := connect.NewProxyConn(nil, nil, nil, zap.NewNop(), connMetrics)
	conn.Claims = &token.GATClaims{}

	got := slices.Sorted(maps.Keys(buildVariables(conn)))
	want := slices.Sorted(slices.Values(template.AllowedWebAppKeys))
	assert.Equal(t, want, got)
}
```

internal/webapphandler/config.go
```diff
@@ -0,0 +1,34 @@
...
+	headers := make(map[string]*template.Template, len(configHeaders))
+
+	for name, value := range configHeaders {
+		tmpl, err := template.New(value)
+		if err != nil {
+			return nil, fmt.Errorf("header %q: %w", name, err)
+		}
+
+		headers[name] = tmpl
+	}
```
minhtule, line 36: This duplicates the validation logic in `internal/config` package. I think we should remove the validation logic there and move it here. In Kubernetes and SSH, we also does some validation in the backend's config.

minhtule (approved):
Fix & merge!

sghiocel (approved):

PR #314, opened 2026-06-02
feat: support wildcard DNS matching for resource addresses

clement0010:
Branched off from #310 

- Issue: https://github.com/Twingate/gateway/issues/309

## Changes

Add RFC 6125 wildcard DNS matching support for resource address validation during CONNECT handling.

- Introduce `matchResourceAddress()` in `internal/connect/connect.go` that supports both exact matching and wildcard patterns (e.g. `*.example.com`)
- Wildcard matching validates a single left-most label only — `*.example.com` matches `api.example.com` but not `example.com` or `foo.bar.example.com`
- Case-insensitive matching for both exact and wildcard patterns
- Rejects invalid patterns: non-leftmost wildcards (`api.*.com`), bare wildcards (`*`), and wildcards without dot separator (`*example.com`)


minhtule (approved):
LGTM! 👍

sghiocel (approved):

PR #315, opened 2026-06-03
test(integration): Add WebApp integration test

clement0010:
Branched out from #310 

## Related Tickets

Issue: #309

## Changes

- Add end-to-end integration test (`TestWebApp`) for the WebApp proxying flow on port `8450`
  - Add `WebAppUser` type in `testutil/user.go` following existing `User`/`SSHUser` patterns
  - Add echo HTTP server utility (`testutil/webapp_server.go`) as the upstream target
- Make `fake.NewClient` accept `token.GeoIPLocation` parameter instead of hardcoding it
  - Update all existing `fake.NewClient` callers with empty `GeoIPLocation{}`

## Notes

The test validates:
- Authentication via CONNECT tunnel with `WEB_APP` resource type
- Header injection via templates (username, groups, JWT, geo-IP headers)
- HTTP reverse proxying to upstream (plain HTTP, no inner TLS upgrade)
- Audit logging via `AssertLogsForREST`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #316, opened 2026-06-03
feat: Add WebApp client to local dev environment

clement0010:
Branched out from #315 

## Related Tickets

Issue: #309

## Changes

- Add WebApp echo server (`tools/local/web_app.go`) that returns method, path, and headers as JSON, and echoes request headers back as response headers
- Add WebApp fake client to the local dev tooling with geo-IP location data
- Add WebApp config (header templates) to the local gateway config
- Update dev environment output with WebApp client address and curl instructions

## Notes

After starting the gateway, test WebApp header forwarding with:

```bash
curl http://<webapp-client-address>
```


minhtule (approved):
LGTM 👍

Issue #146, opened 2025-10-14
chore: Update GoReleaser manfest to use new docker


ekampf:
Resolve this warning:
```
dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info
```

minhtule:
We should also use the the `DOCKER_BUILDX_CACHE` defined in the Makefile. See https://github.com/Twingate/kubernetes-access-gateway/pull/176#discussion_r2669384342

PR #321, opened 2026-06-05
chore: Migrate goreleaser from dockers+docker_manifests to dockers_v2

clement0010:
## Related Tickets

Issue : #146

## Changes

Migrate `.goreleaser.yaml` from the deprecated `dockers` + `docker_manifests` configuration to `dockers_v2` (available since goreleaser v2.12). This leverages Docker buildx native multi-arch manifest support, eliminating ~170 lines of repetitive YAML.

See [deprecation notice](https://goreleaser.com/resources/deprecations/#dockers) 

### What changed

- **`.goreleaser.yaml`**: Replaced `dockers` (4 entries) + `docker_manifests` (16 entries) with `dockers_v2` (4 entries)
- **`Dockerfile.goreleaser`**: Added `ARG TARGETPLATFORM` and updated `COPY` to use platform-prefixed binary path (`$TARGETPLATFORM/gateway`)
- **`Dockerfile.goreleaser-debug`**: Added `ARG TARGETPLATFORM` in final stage, updated `COPY` path, bumped delve to `v1.26.3`
- **`Makefile`**: Renamed builder, split `DOCKER_BUILDX_CACHE` into `DOCKER_BUILDX_CACHE_FROM`/`DOCKER_BUILDX_CACHE_TO`, removed unused `GOLANG_VERSION` and `IMAGE_REGISTRY` env vars from build target
- **`.github/workflows/ci.yaml`**: Split `DOCKER_BUILDX_CACHE` env var into `DOCKER_BUILDX_CACHE_FROM` and `DOCKER_BUILDX_CACHE_TO`

#### Snapshot builds (local / CI)

| Type | Old (`dockers`) | New (`dockers_v2`) |
|------|----------------|-------------------|
| Per-platform tag | `gateway:0.17.0-dev-abc1234-linux-amd64` | `gateway:0.17.0-dev-abc1234-amd64` |
| Per-platform debug tag | `gateway:0.17.0-dev-abc1234-linux-amd64-debug` | `gateway:0.17.0-dev-abc1234-debug-amd64` |
| `linux` prefix | Yes (`-linux-arm64`) | No (`-arm64`) |
| `-debug` suffix position | After platform (`-arm64-debug`) | Before platform (`-debug-arm64`) |

#### Production release (pushed to registry)

| Type | Old (`dockers` + `docker_manifests`) | New (`dockers_v2`) |
|------|-------------------------------------|-------------------|
| Version tag | `gateway:0.17.0` (manifest) + `gateway:0.17.0-linux-amd64` + `gateway:0.17.0-linux-arm64` (per-platform) | `gateway:0.17.0` (multi-arch manifest only) |
| Latest tag | `gateway:latest` (manifest) | `gateway:latest` (multi-arch manifest) |
| Major tag | `gateway:0` (manifest) | `gateway:0` (multi-arch manifest) |
| Minor tag | `gateway:0.17` (manifest) | `gateway:0.17` (multi-arch manifest) |
| Debug version tag | `gateway:0.17.0-debug` (manifest) + per-platform tags | `gateway:0.17.0-debug` (multi-arch manifest only) |
| Per-platform tags in registry | Yes (e.g. `gateway:0.17.0-linux-amd64`) | No — embedded in manifest |

**Key difference**: In the old config, per-platform tags (e.g. `gateway:0.17.0-linux-amd64`) were pushed to the registry alongside manifests. With `dockers_v2`, only multi-arch manifest tags are pushed — per-platform images are embedded in the manifest, not separately tagged.

## Notes

- Requires goreleaser v2.12+ (CI uses `~> v2` which resolves to latest v2.x)
- `docker_manifests` section is fully eliminated — `dockers_v2` handles manifest creation automatically
- Cache flags (`--cache-from`, `--cache-to`) are wired up via env vars, defaulting to empty (no-op) for local builds

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -66,7 +66,6 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          IMAGE_REGISTRY: ghcr.io/twingate
           DOCKER_BUILDX_BUILDER: twingate-gateway-builder
 
       - name: Helm Publish (Github)
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -69,8 +69,8 @@ dockers_v2:
     flags:
       - "--pull"
       - "--builder={{ .Env.DOCKER_BUILDX_BUILDER }}"
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_FROM" }}--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}{{ end }}'
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_TO" }}--cache-to={{ .Env.DOCKER_BUILDX_CACHE_TO }}{{ end }}'
+      - '--cache-from={{ envOrDefault "DOCKER_BUILDX_CACHE_FROM" "" }}'
+      - '--cache-to={{ envOrDefault "DOCKER_BUILDX_CACHE_TO" "" }}'
     extra_files:
       - LICENSE
   # Release: latest/dev, major, minor tags
@@ -97,8 +97,8 @@ dockers_v2:
     flags:
       - "--pull"
       - "--builder={{ .Env.DOCKER_BUILDX_BUILDER }}"
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_FROM" }}--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}{{ end }}'
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_TO" }}--cache-to={{ .Env.DOCKER_BUILDX_CACHE_TO }}{{ end }}'
+      - '--cache-from={{ envOrDefault "DOCKER_BUILDX_CACHE_FROM" "" }}'
+      - '--cache-to={{ envOrDefault "DOCKER_BUILDX_CACHE_TO" "" }}'
     extra_files:
       - LICENSE
   # Debug snapshot + release: version tag only
@@ -119,8 +119,8 @@ dockers_v2:
     flags:
       - "--pull"
       - "--builder={{ .Env.DOCKER_BUILDX_BUILDER }}"
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_FROM" }}--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}{{ end }}'
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_TO" }}--cache-to={{ .Env.DOCKER_BUILDX_CACHE_TO }}{{ end }}'
+      - '--cache-from={{ envOrDefault "DOCKER_BUILDX_CACHE_FROM" "" }}'
+      - '--cache-to={{ envOrDefault "DOCKER_BUILDX_CACHE_TO" "" }}'
     build_args:
       GOLANG_VERSION: "{{ .Env.GOLANG_VERSION }}"
     extra_files:
@@ -149,8 +149,8 @@ dockers_v2:
     flags:
       - "--pull"
       - "--builder={{ .Env.DOCKER_BUILDX_BUILDER }}"
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_FROM" }}--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}{{ end }}'
-      - '{{ if isEnvSet "DOCKER_BUILDX_CACHE_TO" }}--cache-to={{ .Env.DOCKER_BUILDX_CACHE_TO }}{{ end }}'
+      - '--cache-from={{ envOrDefault "DOCKER_BUILDX_CACHE_FROM" "" }}'
+      - '--cache-to={{ envOrDefault "DOCKER_BUILDX_CACHE_TO" "" }}'
     build_args:
       GOLANG_VERSION: "{{ .Env.GOLANG_VERSION }}"
     extra_files:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -69,8 +69,8 @@ dockers_v2:
     flags:
       - "--pull"
       - "--builder={{ .Env.DOCKER_BUILDX_BUILDER }}"
-      - "--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}"
-      - "--cache-to={{ .Env.DOCKER_BUILDX_CACHE_TO }}"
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -2,10 +2,11 @@ GOLANG_VERSION 		?= $(shell cat .tool-versions | grep golang | cut -d' ' -f2)
...
 VERSION 			?= $(shell go tool svu current)
 REGISTRY 			?= twingate
-IMAGE				:= kubernetes-gateway
+IMAGE				:= gateway
 IMAGE_NAME			:= $(REGISTRY)/$(IMAGE)
-DOCKER_BUILDX_BUILDER ?= twingate-kubernetes-gateway-builder
-DOCKER_BUILDX_CACHE ?=
+DOCKER_BUILDX_BUILDER ?= twingate-gateway-builder
+DOCKER_BUILDX_CACHE_FROM ?=
+DOCKER_BUILDX_CACHE_TO ?=
```
clement0010, line 9:
```
DOCKER_BUILDX_CACHE_FROM ?= type=local,src=/tmp/.buildx-cache                       
DOCKER_BUILDX_CACHE_TO ?= type=local,dest=/tmp/.buildx-cache
```

I tried using local cache, but the performance gain is negligible. I think our Dockerfile is too simple to take advantage of the cache. 

The first build is also slower because it needs to write the cache to the directory.

Makefile
```diff
@@ -127,7 +128,7 @@ prepare-buildx: ##@build Prepare buildx
 
 .PHONY: build
 build: prepare-buildx ##@build Build the Go binaries and container images
-	DOCKER_BUILDX_BUILDER=$(DOCKER_BUILDX_BUILDER) GOLANG_VERSION=$(GOLANG_VERSION) IMAGE_REGISTRY=$(REGISTRY) goreleaser release --snapshot --clean
```
clement0010, line 130: Removing `GOLANG_VERSION` and `IMAGE_REGISTRY` because they are not used in `goreleaser` at all

.goreleaser.yaml
```diff
@@ -10,7 +10,7 @@ before:
     - go mod tidy
 
 snapshot:
-  version_template: '{{ .Version }}-{{envOrDefault "VERSION_PRERELEASE" "local"}}+{{ .ShortCommit }}'
+  version_template: '{{ .Version }}-{{envOrDefault "VERSION_PRERELEASE" "local"}}-{{ .ShortCommit }}'
```
clement0010, line 13: I think we can update this so we can avoid doing `{{ replace .Version \"+\" \"-\" }}` everywhere.

.goreleaser.yaml
```diff
@@ -49,173 +49,109 @@ builds:
...
+    images:
+      - "twingate/{{ .ProjectName }}"
+      - "ghcr.io/twingate/{{ .ProjectName }}"
+    tags:
+      # Example: twingate/gateway:latest (or dev)
+      - "{{ if not .Prerelease }}latest{{ else }}dev{{ end }}"
+      # Example: twingate/gateway:0
+      - "{{ if not .Prerelease }}{{.Major}}{{ end }}"
+      # Example: twingate/gateway:0.2
+      - "{{ if not .Prerelease }}{{.Major}}.{{.Minor}}{{ end }}"
```
clement0010, line 91: This `.Prerelease` check is equivalent to setting `skip_push: auto` in `docker_manifests` in the old config. From the [docs](https://goreleaser.com/customization/package/docker_manifest/):

> If set to `auto`, the manifest will not be created in case there is an indicator of a prerelease in the tag, e.g. v1.0.0-rc1.

.goreleaser.yaml
```diff
@@ -49,173 +49,109 @@ builds:
...
+    flags:
       - "--pull"
       - "--builder={{ .Env.DOCKER_BUILDX_BUILDER }}"
-      - "--platform=linux/arm64"
-      - "--label=org.opencontainers.image.created={{.Date}}"
-      - "--label=org.opencontainers.image.title={{.ProjectName}}"
-      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
-      - "--label=org.opencontainers.image.version={{.Tag}}"
+      - "--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}"
+      - "--cache-to={{ .Env.DOCKER_BUILDX_CACHE_TO }}"
```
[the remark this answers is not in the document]
clement0010: An empty value is ignored by `Docker` so i think it is fine here

Makefile
minhtule: I think we use `GOLANG_VERSION` in the [debug build](https://github.com/Twingate/gateway/blob/8d5619d397746d63815dea9a2ba921d2916773e9/Dockerfile.goreleaser-debug#L5)

.goreleaser.yaml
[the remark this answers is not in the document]
minhtule: Does Goreleaser work if the envvar doesn't exist? If so, we need to add these envvars to the `release.yaml`

test/e2e/e2e_test.go
```diff
@@ -136,7 +136,7 @@ func downloadCaddyCACert(t *testing.T) string {
 func getDockerImageTag(t *testing.T) string {
 	t.Helper()
 
-	var imageReference = "twingate/gateway:*-linux-"
+	var imageReference = "twingate/gateway:*-"
```
minhtule, line 139: This might use debug image. We probably need filter to filter the debug image manually 🤔

minhtule (approved):
Fix & merge!

.github/workflows/release.yaml
```diff
@@ -66,7 +66,6 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          IMAGE_REGISTRY: ghcr.io/twingate
```
clement0010, line 69: This envvar is not used in goreleaser, we always build `ghcr.io/twingate` and `twingate` registry

.goreleaser.yaml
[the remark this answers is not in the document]
clement0010: > Does Goreleaser work if the envvar doesn't exist?

Nope, we need to pass in those envvars. In that case, I think it'd be easier to implement the conditional check 

```yaml
{{ if isEnvSet "DOCKER_BUILDX_CACHE_FROM" }}--cache-from={{ .Env.DOCKER_BUILDX_CACHE_FROM }}{{ end }}
```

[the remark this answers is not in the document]
minhtule: Or we can use `envOrDefault` template function.
```
- "--cache-from={{ envOrDefault \"DOCKER_BUILDX_CACHE_FROM\" \"\" }}"
- "--cache-to={{ envOrDefault \"DOCKER_BUILDX_CACHE_TO\" \"\" }}"
```

ekampf (approved):

PR #347, opened 2026-06-28
fix: Use `rsa2048` as the default AWS signature type

minhtule:

- Issue: #346 

## Changes

- Default the AWS signature type to `rsa2048` (SHA-256) when signatureType is unset.
- Emit a deprecation warning when `pkcs7` is explicitly configured.


clement0010 (approved):
LGTM 👍🏻

internal/sshhandler/vault.go
```diff
@@ -95,16 +95,15 @@ func newAWSAuthMethod(awsConfig *gatewayconfig.SSHCAVaultAWSConfig) (*aws.AWSAut
...
 		opts = append(opts, aws.WithIdentitySignature())
-	case "rsa2048":
-		opts = append(opts, aws.WithRSA2048Signature())
 	case "pkcs7":
+		logger.Warn("Vault AWS EC2 auth signatureType 'pkcs7' is deprecated as it relies on SHA-1; use 'rsa2048' instead")
+
 		opts = append(opts, aws.WithPKCS7Signature())
 	default:
-		// Use Vault SDK default (pkcs7)
+		opts = append(opts, aws.WithRSA2048Signature())
```
sghiocel, line 106: nit: Might be useful to add a `logger.Info` message here, to log that the default was chosen - `rsa2048`

sghiocel (approved):
LGTM - Check + decide on the minor comment, otherwise good to merge!

internal/sshhandler/vault.go
minhtule: Good point, though I'd lean towards not logging it here: it'd fire on every normal startup, and we don't log other applied defaults either (e.g. the Gateway key type defaulting to `ed25519`, cert TTLs, ports). If we want visibility into effective config, maybe we revisit it more consistently across all defaults, likely at debug. WDYT?

PR #351, opened 2026-06-29
fix: Split SSH gateway key into separate host and user keys

minhtule:

- Closes #350

## Changes

- Generate separate host and user keypairs instead of one shared gateway key

### DIFF
--- a/internal/sshhandler/config.go
+++ b/internal/sshhandler/config.go
@@ -28,7 +28,6 @@ var (
 
 // Default values for SSH configuration.
 const (
-	defaultKeyType     = keyTypeED25519
 	defaultHostCertTTL = 24 * time.Hour
 	defaultUserCertTTL = 5 * time.Minute
 )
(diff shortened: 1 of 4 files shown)


internal/sshhandler/config.go
```diff
@@ -78,12 +83,19 @@ func NewConfig(auditLogConfig *config.AuditLogConfig, sshCfg *config.SSHConfig,
 		keyType = defaultKeyType
 	}
 
-	gatewaySigner, gatewayPublicKey, err := keyConfig{
+	keyCfg := keyConfig{
 		typ:  keyType,
 		bits: sshCfg.Gateway.Key.Bits,
-	}.Generate(rand.Reader)
+	}
```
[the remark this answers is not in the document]
minhtule: Fixed in https://github.com/Twingate/gateway/pull/351/commits/2008c11f2e14b31560b0aa0566f995c7c8f4b62a

internal/sshhandler/config.go
```diff
@@ -72,18 +76,19 @@ func NewConfig(auditLogConfig *config.AuditLogConfig, sshCfg *config.SSHConfig,
...
+
+	hostSigner, hostPublicKey, err := keyCfg.Generate(rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate gateway host key: %w", err)
 	}
 
-	gatewaySigner, gatewayPublicKey, err := keyConfig{
-		typ:  keyType,
-		bits: sshCfg.Gateway.Key.Bits,
-	}.Generate(rand.Reader)
```
minhtule, line 84: There is an existing bug that is also fixed with this change: the old code constructs `keyConfig` instance directly instead of using the constructor [`newKeyConfig`](https://github.com/Twingate/gateway/blob/6bf74c45e7fcabc44b2afe6603ce53f83ff730f8/internal/sshhandler/keys.go#L37-L78). As such, key type in SSH format like `ssh-ed25519` is treated as error.

internal/sshhandler/config.go
```diff
@@ -72,18 +76,19 @@ func NewConfig(auditLogConfig *config.AuditLogConfig, sshCfg *config.SSHConfig,
 		return nil, fmt.Errorf("failed to create ca: %w", err)
 	}
 
-	// Apply defaults for Gateway's key config
-	keyType := keyType(sshCfg.Gateway.Key.Type)
-	if keyType == "" {
-		keyType = defaultKeyType
```
minhtule, line 78: Move this default to `newKeyConfig` constructor because it's more appropriate and also testable.

clement0010 (approved):
LGTM 👍🏻

sghiocel (approved):

PR #361, opened 2026-06-30
fix: Validate SSH certificate that it doesn't contain more privilege than requested

minhtule:

- Closes #360

## Changes

- Verify a CA-issued SSH certificate grants no more than the signing request before use, and reject it otherwise.


internal/sshhandler/ca.go
```diff
@@ -288,5 +289,55 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
+	for _, p := range cert.ValidPrincipals {
+		if _, ok := requestedPrincipals[p]; !ok {
+			return fmt.Errorf("%w: unexpected principal %q", errCertPolicyViolation, p)
+		}
+	}
+
+	maxValidBefore := mustUint64(time.Now().Add(req.ttl).Add(clockSkewBuffer))
+	if cert.ValidBefore > maxValidBefore {
+		return fmt.Errorf("%w: validity %d exceeds requested TTL (max %d)", errCertPolicyViolation, cert.ValidBefore, maxValidBefore)
+	}
```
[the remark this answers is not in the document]
minhtule: The TTL is anchored at when the gateway receives and starts using the cert (it's used immediately on return), not when the request was sent. Treating "valid for `ttl` from when the cert becomes usable" as the intended meaning is reasonable, and it's also what Vault does (`ValidBefore = signing_time + ttl`). Under that reading the response delay doesn't grant extra lifetime, so I'll leave the check as-is

internal/sshhandler/ca_test.go
```diff
@@ -274,3 +280,346 @@ func TestMustUint64_PanicsOnNegativeTime(t *testing.T) {
...
+	ca := newVaultTestCA(t, func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusInternalServerError)
+	})
+
+	_, err = ca.sign(context.Background(), req)
+	require.Error(t, err)
+	require.NotErrorIs(t, err, errVaultSignFailed)
+}
+
+func TestVerifyCertificate(t *testing.T) {
```
minhtule, line 515: Only this test is needed for the code change. The other tests for Vault CA (which didn't have any test!) is to improve code coverage.

internal/sshhandler/ca.go
```diff
@@ -288,5 +289,75 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
```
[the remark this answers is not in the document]
minhtule: Improve the cert renewal logic in https://github.com/Twingate/gateway/pull/362 so backdating `ValidAfter` doesn't cause any issue. Enforcing a lower bound is kind of arbitrary so not adding this validation.

internal/sshhandler/ca_test.go
```diff
@@ -274,3 +280,321 @@ func TestMustUint64_PanicsOnNegativeTime(t *testing.T) {
...
+		}, wantErrMsg: "different public key"},
+		{name: "extra principal", req: userReq, setupFn: func(c *ssh.Certificate) {
+			c.ValidPrincipals = []string{"alice", "root"}
+		}, wantErrMsg: "unexpected principal"},
+		{name: "validity exceeds TTL", req: userReq, setupFn: func(c *ssh.Certificate) {
+			c.ValidBefore = mustUint64(now.Add(2 * time.Hour))
+		}, wantErrMsg: "exceeds requested TTL"},
+		{name: "no expiry", req: userReq, setupFn: func(c *ssh.Certificate) {
+			c.ValidBefore = ssh.CertTimeInfinity
+		}, wantErrMsg: "exceeds requested TTL"},
```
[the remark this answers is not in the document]
minhtule: Improve the cert renewal logic in https://github.com/Twingate/gateway/pull/362 so backdating `ValidAfter` doesn't cause any issue. Enforcing a lower bound is kind of arbitrary so not adding this validation.

internal/sshhandler/ca_test.go
```diff
@@ -274,3 +280,346 @@ func TestMustUint64_PanicsOnNegativeTime(t *testing.T) {
...
+		{name: "unparseable public key", responseData: map[string]any{"public_key": "not-a-key"}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ca := newVaultTestCA(t, func(w http.ResponseWriter, _ *http.Request) {
+				_ = json.NewEncoder(w).Encode(map[string]any{"data": tt.responseData})
+			})
+
+			publicKey, err := ca.publicKey(context.Background())
```
clement0010:
```suggestion
			publicKey, err := ca.publicKey(t.Context())
```

internal/sshhandler/ca.go
```diff
@@ -288,5 +289,75 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
+		if granted != requested {
+			return fmt.Errorf("%w: critical option %q granted value %q, requested %q", errCertPolicyViolation, opt, granted, requested)
+		}
+	}
+
+	for opt := range cert.CriticalOptions {
+		if _, ok := req.permissions.CriticalOptions[opt]; !ok {
+			return fmt.Errorf("%w: unexpected critical option %q", errCertPolicyViolation, opt)
+		}
+	}
```
clement0010: If `CriticalOptions` is restrictive, having extra critical options should technically be fine? 🤔

internal/sshhandler/ca_test.go
```diff
@@ -274,3 +280,346 @@ func TestMustUint64_PanicsOnNegativeTime(t *testing.T) {
...
+		})
+	}
+}
+
+func TestVaultCA_PublicKey_RequestFails(t *testing.T) {
+	ca := newVaultTestCA(t, func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusInternalServerError)
+	})
+
+	_, err := ca.publicKey(context.Background())
```
clement0010: ditto

internal/sshhandler/ca_test.go
```diff
@@ -274,3 +280,346 @@ func TestMustUint64_PanicsOnNegativeTime(t *testing.T) {
...
+
+				// The server checks the request payload the gateway built for the request.
+				assert.Equal(t, tt.wantCertType, gotData["cert_type"])
+				assert.Equal(t, string(ssh.MarshalAuthorizedKey(publicKey)), gotData["public_key"])
+				assert.Equal(t, tt.wantPrincipals, gotData["valid_principals"])
+				assert.Equal(t, tt.req.ttl.String(), gotData["ttl"])
+
+				if tt.wantExtensions == nil {
+					_, ok := gotData["extensions"]
+					assert.False(t, ok, "extensions must not be sent for host certs")
```
clement0010: nit:
```suggestion
					assert.NotContains(t, gotData, "extensions", "extensions must not be sent for host certs")

```

clement0010 (approved):
Fix & merge! One question: see below

internal/sshhandler/ca.go
minhtule: That's true in general but `force-command` is the exception. It substitutes what runs on the session rather than narrowing it, so a CA injecting it (we currently request zero critical options) would silently change behavior. This is a fail-closed check on the CA doing exactly what we asked, so I'd keep the exact-match and reject anything unrequested.

internal/sshhandler/ca.go
```diff
@@ -288,5 +289,75 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
+	if !keysEqual(cert.Key, req.publicKey) {
+		return fmt.Errorf("%w: certificate is bound to a different public key", errCertPolicyViolation)
+	}
+
+	requestedPrincipals := make(map[string]struct{}, len(req.principals))
+	for _, p := range req.principals {
+		requestedPrincipals[p] = struct{}{}
+	}
+
+	for _, p := range cert.ValidPrincipals {
```
sghiocel: Here we do no validation if `cert.ValidPrincipals` is empty. So if a cert with no principals is returned, then it will become overly permissive, see:

https://github.com/golang/crypto/blob/45460e079737ecb64f30d79d3d6fc2914494fa66/ssh/certs.go#L421-L433

We should have a similar check to ensure this doesn't happen

minhtule: Good point! Updated in https://github.com/Twingate/gateway/pull/361/commits/e4c0449e915238a49414d499790c85a1851b06aa to match the `principals` list as requested.

internal/sshhandler/ca.go
```diff
@@ -288,5 +290,66 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
+	return nil
+}
+
+func principalSet(principals []string) map[string]struct{} {
+	set := make(map[string]struct{}, len(principals))
+	for _, p := range principals {
+		set[p] = struct{}{}
+	}
+
+	return set
```
minhtule, line 354: At some point, we might want to create `internal/libs/set.go` for this 🙈

internal/sshhandler/ca.go
```diff
@@ -288,5 +290,66 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
+	}
+
+	if !keysEqual(cert.Key, req.publicKey) {
+		return fmt.Errorf("%w: certificate is bound to a different public key", errCertPolicyViolation)
+	}
+
+	// Require exact equality: an empty principals list grants every principal, and any extra
+	// principal grants access the request never asked for.
+	if !maps.Equal(principalSet(cert.ValidPrincipals), principalSet(req.principals)) {
+		return fmt.Errorf("%w: granted principals %q do not match requested %q", errCertPolicyViolation, cert.ValidPrincipals, req.principals)
```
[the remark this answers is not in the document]
minhtule: There is no such error. `%q` would double quote the slice of strings correctly. I think it's nicer to have quotes here to separate the elements.
```go
a := []string{"alice", "bob", "charlie"}
fmt.Printf("List %v, %q", a, a)

# Output
# List [alice bob charlie], ["alice" "bob" "charlie"]
```

internal/sshhandler/ca.go
```diff
@@ -288,5 +290,66 @@ func (ca *vaultCA) sign(ctx context.Context, req *certificateRequest) (*ssh.Cert
...
+
+	maxValidBefore := mustUint64(time.Now().Add(req.ttl).Add(clockSkewBuffer))
+	if cert.ValidBefore > maxValidBefore {
+		return fmt.Errorf("%w: validity %d exceeds requested TTL (max %d)", errCertPolicyViolation, cert.ValidBefore, maxValidBefore)
+	}
+
+	// Critical options are restrictions, so dropping one widens privilege; require the
+	// cert's options to match the request exactly.
+	if !maps.Equal(cert.CriticalOptions, req.permissions.CriticalOptions) {
+		return fmt.Errorf("%w: granted critical options %q do not match requested %q", errCertPolicyViolation, cert.CriticalOptions, req.permissions.CriticalOptions)
```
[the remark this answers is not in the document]
minhtule: https://github.com/Twingate/gateway/pull/361#discussion_r3562757317

sghiocel (approved):

PR #362, opened 2026-07-01
fix: Make SSH certificate auto-renewal robust to backdated validity

minhtule:

- Issue: N/A

## Changes

- Measure renewal time from now when a cert's `ValidAfter` is backdated, so an already-started validity window doesn't drag the schedule into the past
- Floor the renewal timer delay at `retryInterval`
  - A backdated or already-expired cert yields a renewal time in the past, so without a floor the timer fires immediately and re-signs in a tight loop, hammering the CA


internal/sshhandler/cert.go
```diff
@@ -155,8 +155,19 @@ func renewTime(cert *ssh.Certificate) time.Time {
...
 	lifetime := expiresAt.Sub(issuedAt)
 
 	return issuedAt.Add(time.Duration(float64(lifetime) * renewFraction))
 }
+
+// renewalDelay returns the delay until t, floored at retryInterval so a renewal time in the past
+// (e.g. from a backdated certificate) cannot busy-loop the timer.
+func renewalDelay(t time.Time) time.Duration {
+	return max(time.Until(t), retryInterval)
+}
```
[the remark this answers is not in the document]
minhtule: The floor only affects renewals under ~12.5s out (0.8 × lifetime < retryInterval). The shortest configured TTL is 5m, which renews ~240s out, so a legitimate near-term renewal is never delayed in practice. The floor is also an intentional debounce: it caps re-signs at one per retryInterval, matching the error-retry path and preventing CA hammering if a CA keeps issuing near-expired certs. Clamping only non-positive durations would remove that guard.

clement0010: If for some reasons, the Vault CA kept returning a backdated certificate, maybe we want to consider adding a log here so we can detect it? 🤔

clement0010 (approved):
LGTM 👍🏻  One comment

internal/sshhandler/cert.go
minhtule: Added in https://github.com/Twingate/gateway/pull/362#discussion_r3506468758

clement0010 (approved):
LGTM 👍🏻

minhtule:
We're changing how the host certificate is being renewed (lazily when it expires) so this change would become unnecessary.

PR #363, opened 2026-07-01
feat: Support GAT port mapping and resource alias translation on CONNECT

clement0010:

## Changes

The controller issues per-resource routing details in the GAT — port mapping under `resource.gateway_metadata` and alternate names under `resource.aliases`. On CONNECT, the Gateway now resolves the destination against both before forwarding:

- **Address / alias**: the CONNECT host must match the token's `resource.address` (exact or RFC 6125 wildcard) or any of its `resource.aliases` (exact match); a matched alias is translated to the upstream resource address. Anything else is refused with 400.
- **Downstream port** (`downstream.port`, client-facing): the CONNECT port must equal the token's downstream port, otherwise the connection is refused with 400.
- **Upstream port** (`upstream.port`, backend-facing): once the address and downstream-port checks pass, the target is rewritten to the resource address and upstream port before forwarding to the backend.


internal/token/gat_claims.go
```diff
@@ -111,9 +131,26 @@ const (
...
-	ID      string       `json:"id"`
-	Type    ResourceType `json:"type"`
-	Address string       `json:"address"`
+	ID              string          `json:"id"`
+	Type            ResourceType    `json:"type"`
+	Address         string          `json:"address"`
+	GatewayMetadata GatewayMetadata `json:"gateway_metadata"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
+}
+
+// GatewayMetadata carries per-resource routing details from the GAT.
```
minhtule: Gateway metadata could contain other information than just for routing.

internal/token/gat_claims.go
```diff
@@ -111,9 +131,26 @@ const (
...
+type GatewayMetadata struct {
+	Downstream Downstream `json:"downstream"`
+	Upstream   Upstream   `json:"upstream"`
+}
+
+// Downstream is the client-facing endpoint the CONNECT destination must target.
+type Downstream struct {
+	Port int `json:"port"`
+}
```
minhtule: - `Downstream` describes the connection between the protocol client and the Gateway
- `Port` - the port that the protocol client connects to

internal/token/gat_claims.go
```diff
@@ -111,9 +131,26 @@ const (
...
+
+// Upstream is the backend endpoint the CONNECT destination is forwarded to.
+type Upstream struct {
+	Port int `json:"port"`
```
minhtule: - `Upstream` describes the connection between the Gateway and the upstream resource
- `Port` - the port on the upstream resource that the Gateway forwards the connection to

internal/connect/connect.go
```diff
@@ -162,6 +171,31 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
 	}, nil
 }
 
+// resolveUpstreamAddress verifies the CONNECT destination port matches the GAT
+// downstream port and maps it to the upstream port for backend forwarding. Both
+// ports are validated for presence and range when the token is parsed, so they
+// are guaranteed to be within the valid range here.
```
minhtule: I think the 2nd sentence is redundant. This function could translate the port without any validation.

internal/token/gat_claims.go
```diff
@@ -111,9 +131,26 @@ const (
 )
 
 type Resource struct {
-	ID      string       `json:"id"`
-	Type    ResourceType `json:"type"`
-	Address string       `json:"address"`
+	ID              string          `json:"id"`
+	Type            ResourceType    `json:"type"`
+	Address         string          `json:"address"`
+	GatewayMetadata GatewayMetadata `json:"gateway_metadata"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
```
minhtule, line 138: Let's make sure all GATs include the `gateway_metadata` before merging.

minhtule (approved):
LGTM 👍

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+			Message: fmt.Sprintf("failed to parse CONNECT destination: %v", err),
+			Err:     err,
+		}
+	}
+
+	switch {
+	case matchResourceAddress(resource.Address, host):
+		//	Keep existing upstream address
+	case matchResourceAliases(resource.Aliases, host):
+		//	Rewrite upstream address to GAT token address
```
minhtule: The comment is using tabs after `//`?

minhtule: While Twingate resource's address could only be host, address in general includes both host and port. These comments could be clearer e.g.
```go
switch {
case matchResourceAddress(resource.Address, host):
	// host matches the resource address; forward to the same host
case matchResourceAliases(resource.Aliases, host):
	// host matches an alias; forward to the resource's address
	host = resource.Address
```

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+		host = resource.Address
+	default:
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("CONNECT destination address %s does not match token resource address %s", host, resource.Address),
```
minhtule: We should also include the alias (or not include address)
```go
Message: fmt.Sprintf("CONNECT host %s does not match resource address %s or aliases %v", host, resource.Address, resource.Aliases),
```

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
 }
 
+// resolveUpstreamAddress verifies the CONNECT destination against the GAT
+// resource and maps it to the upstream address for backend forwarding.
+func resolveUpstreamAddress(address string, resource token.Resource) (string, *HTTPError) {
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("failed to parse CONNECT destination: %v", err),
```
minhtule: I think we should avoid using "destination" to avoid confusion because the Gateway has to translate this address to the actual destination. `CONNECT address` or `CONNECT target` is probably more appropriate.

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+			Message: "CONNECT destination resolves to wildcard upstream address " + host,
+			Err:     nil,
+		}
+	}
+
+	requestedPort, err := strconv.Atoi(port)
+	if err != nil {
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("failed to parse CONNECT destination port: %v", err),
```
minhtule: ditto. dropping "destination" to  "failed to parse CONNECT port"

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+			Err:     err,
+		}
+	}
+
+	metadata := resource.GatewayMetadata
+	if requestedPort != metadata.Downstream.Port {
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("CONNECT destination port %d does not match token downstream port %d", requestedPort, metadata.Downstream.Port),
```
minhtule: Ditto. I would drop "destination"

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("CONNECT destination address %s does not match token resource address %s", host, resource.Address),
+			Err:     nil,
+		}
+	}
+
+	if token.IsWildcardAddress(host) {
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: "CONNECT destination resolves to wildcard upstream address " + host,
```
minhtule: `"CONNECT host resolves to a wildcard: " + host`

internal/connect/connect_test.go
```diff
@@ -312,6 +319,50 @@ func TestConnectValidator_ParseConnect(t *testing.T) {
...
+		signature := c.sign(sigData)
+		req.Header.Set(AuthSignatureHeaderKey, signature)
+		req.Header.Set(ConnIDHeaderKey, "conn-id")
+
+		connectInfo, err := validator.ParseConnect(req, []byte(sigData))
+
+		require.NoError(t, err)
+		assert.Equal(t, "10.0.0.1:8443", connectInfo.Address)
+		assert.Equal(t, "conn-id", connectInfo.ConnID)
+	})
```
minhtule: These 2 tests overlap with test cases in `TestResolveUpstreamAddress`. While we need to make sure the wiring is correct, we shouldn't duplicate all the test cases. I would keep the port mapping and remove the alias test. In the `"Successful authentication"`, I would also assert the address there. Also we should keep the successful test together (moving the port mapping test below "Successful authentication") instead of just appending code at the end.

internal/connect/connect_test.go
```diff
@@ -451,3 +597,61 @@ func TestMatchResourceAddress(t *testing.T) {
...
+		want    bool
+	}{
+		{
+			name:    "exact match",
+			aliases: []string{"app.internal"},
+			host:    "app.internal",
+			want:    true,
+		},
+		{
+			name:    "exact match case insensitive",
```
minhtule: nit: should drop "exact". Matching is either exact or not...

internal/connect/connect.go
```diff
@@ -162,6 +146,61 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+		host = resource.Address
+	default:
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("CONNECT destination address %s does not match token resource address %s", host, resource.Address),
+			Err:     nil,
+		}
+	}
+
+	if token.IsWildcardAddress(host) {
```
minhtule: I think it's better to use the `hostnameRegexp` in the master branch. Checking only for `.*` prefix and treating the value as a wildcard address don't seem right.

clement0010: Updated in [5f51ff1](https://github.com/Twingate/gateway/pull/363/commits/5f51ff1b496c9e733f6c43a59271fc138897c684)

internal/connect/connect_test.go
clement0010: Thanks for the suggestions. I updated the test structure in [e79bbb9](https://github.com/Twingate/gateway/pull/363/commits/e79bbb9d251a2b0658847e7f51d8fdc4ec1d3412)

internal/connect/connect.go
clement0010: > The comment is using tabs after //?

Yeah, I removed it

internal/token/gat_claims.go
```diff
@@ -58,6 +65,20 @@ func (p GATClaims) Validate() error {
 		return fmt.Errorf("%w: %w %q", jwt.ErrTokenInvalidClaims, errUnsupportedVersion, p.Version)
 	}
 
+	if err := validatePort(p.Resource.GatewayMetadata.Downstream.Port, "resource.gateway_metadata.downstream.port"); err != nil {
+		return err
+	}
+
+	return validatePort(p.Resource.GatewayMetadata.Upstream.Port, "resource.gateway_metadata.upstream.port")
```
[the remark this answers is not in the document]
clement0010: All GAT token should have both `downstream` and `upstream` fields in gateway_metadata

minhtule (approved):
LGTM 👍

sghiocel (approved):

PR #365, opened 2026-07-01
feat: Support GAT `request_header_rewrites` for Web App traffic

clement0010:

## Changes

For Web App resources, the controller can specify per-resource `request_header_rewrites` in the GAT under `resource.gateway_metadata`. This PR makes the Gateway apply those header rewrites to Web App requests before forwarding them upstream.

### DIFF
--- a/internal/token/gat_claims.go
+++ b/internal/token/gat_claims.go
@@ -111,9 +111,15 @@ const (
 )
 
 type Resource struct {
-	ID      string       `json:"id"`
-	Type    ResourceType `json:"type"`
-	Address string       `json:"address"`
+	ID              string          `json:"id"`
+	Type            ResourceType    `json:"type"`
+	Address         string          `json:"address"`
+	GatewayMetadata GatewayMetadata `json:"gateway_metadata"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
+}
+
+// GatewayMetadata carries per-resource routing details from the GAT.
+type GatewayMetadata struct {
+	RequestHeaderRewrites map[string]string `json:"request_header_rewrites,omitempty"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
 }
 
 // PublicKey is a wrapper for ecdsa.PublicKey that adds support for JSON
--- a/internal/token/gat_claims.go
+++ b/internal/token/gat_claims.go
@@ -117,7 +117,6 @@ type Resource struct {
 	GatewayMetadata GatewayMetadata `json:"gateway_metadata"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
 }
 
-// GatewayMetadata carries per-resource routing details from the GAT.
 type GatewayMetadata struct {
 	RequestHeaderRewrites map[string]string `json:"request_header_rewrites,omitempty"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
 }
--- a/internal/webapphandler/handler.go
+++ b/internal/webapphandler/handler.go
@@ -81,16 +81,21 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
 	}
 
 	// Per-resource request header rewrites from the GAT are applied last, so they override
-	// any config headers with the same name.
+	// any config headers with the same name. Malformed or unsupported headers are skipped
+	// rather than failing the request.
 	for headerName, value := range conn.GATClaims().Resource.GatewayMetadata.RequestHeaderRewrites {
 		tmpl, err := template.New(value)
 		if err != nil {
-			return fmt.Errorf("header %q: %w", headerName, err)
+			conn.Logger.Debug("skipping GAT request header rewrite", zap.String("header", headerName), zap.String("value", value), zap.Error(err))
+
+			continue
 		}
 
 		headerValue, err := tmpl.Evaluate(variables)
 		if err != nil {
-			return fmt.Errorf("header %q: %w", headerName, err)
+			conn.Logger.Debug("skipping GAT request header rewrite", zap.String("header", headerName), zap.String("value", value), zap.Error(err))
+
+			continue
 		}
 
 		r.Out.Header.Set(headerName, headerValue)
--- a/internal/webapphandler/handler.go
+++ b/internal/webapphandler/handler.go
@@ -80,5 +80,21 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
 		r.Out.Header.Set(headerName, headerValue)
 	}
 
+	// Per-resource request header rewrites from the GAT are applied last, so they override
+	// any config headers with the same name.
+	for headerName, value := range conn.GATClaims().Resource.GatewayMetadata.RequestHeaderRewrites {
+		tmpl, err := template.New(value)
+		if err != nil {
+			return fmt.Errorf("header %q: %w", headerName, err)
+		}
+
+		headerValue, err := tmpl.Evaluate(variables)
+		if err != nil {
+			return fmt.Errorf("header %q: %w", headerName, err)
+		}
+
+		r.Out.Header.Set(headerName, headerValue)
+	}
+
 	return nil
 }
--- a/internal/webapphandler/handler.go
+++ b/internal/webapphandler/handler.go
@@ -86,14 +86,14 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
 	for headerName, value := range conn.GATClaims().Resource.GatewayMetadata.RequestHeaderRewrites {
 		tmpl, err := template.New(value)
 		if err != nil {
-			conn.Logger.Debug("skipping GAT request header rewrite", zap.String("header", headerName), zap.String("value", value), zap.Error(err))
+			conn.Logger.Warn("skipping GAT request header rewrite", zap.String("header", headerName), zap.Error(err))
 
 			continue
 		}
 
 		headerValue, err := tmpl.Evaluate(variables)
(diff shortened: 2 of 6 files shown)


internal/webapphandler/handler.go
```diff
@@ -80,5 +80,21 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
 		r.Out.Header.Set(headerName, headerValue)
 	}
 
+	// Per-resource request header rewrites from the GAT are applied last, so they override
+	// any config headers with the same name.
+	for headerName, value := range conn.GATClaims().Resource.GatewayMetadata.RequestHeaderRewrites {
+		tmpl, err := template.New(value)
+		if err != nil {
+			return fmt.Errorf("header %q: %w", headerName, err)
```
clement0010: This has the same error as above in line 77. We could set the panic message to something like `header from token ...` but not sure if it'd be useful 🤔

internal/webapphandler/handler.go
```diff
@@ -80,5 +80,21 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
 		r.Out.Header.Set(headerName, headerValue)
 	}
 
+	// Per-resource request header rewrites from the GAT are applied last, so they override
+	// any config headers with the same name.
+	for headerName, value := range conn.GATClaims().Resource.GatewayMetadata.RequestHeaderRewrites {
+		tmpl, err := template.New(value)
+		if err != nil {
```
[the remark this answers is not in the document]
clement0010: The panic is handled in [`auditlog_middleware`](https://github.com/Twingate/gateway/blob/16b6a1a1432539f225d25aa4832daed573be0b50/internal/httpproxy/audit_middleware.go#L114-L123)

internal/token/gat_claims.go
```diff
@@ -111,9 +111,15 @@ const (
...
-	ID      string       `json:"id"`
-	Type    ResourceType `json:"type"`
-	Address string       `json:"address"`
+	ID              string          `json:"id"`
+	Type            ResourceType    `json:"type"`
+	Address         string          `json:"address"`
+	GatewayMetadata GatewayMetadata `json:"gateway_metadata"` //nolint:tagliatelle // GAT wire format from the controller uses snake_case
+}
+
+// GatewayMetadata carries per-resource routing details from the GAT.
```
minhtule: The metadata could contain information that are not for routing like this request header. I think this comment is not necessary.

internal/webapphandler/handler.go
minhtule: yeah, I would make the error message slightly different like you suggested or "token header ..."

internal/webapphandler/handler_test.go
```diff
@@ -110,6 +117,39 @@ func TestRewrite(t *testing.T) {
...
+			}),
+			headers: map[string]string{
+				"X-Config": "{{groups}}",
+			},
+			wantHeaders: map[string]string{
+				"X-Config":       "Everyone,Engineering",
+				"X-GAT-Static":   "static-value",
+				"X-GAT-Username": "alice@acme.com",
+				"X-GAT-Auth":     "Bearer test-token",
+			},
```
minhtule, line 148: Would be more robust to have a header that is defined in the config but is overridden by the token.

test/integration/web_app_test.go
```diff
@@ -108,12 +112,15 @@ func TestWebApp(t *testing.T) {
...
-		"X-Twingate-Username":           "alex@acme.com",
+		// From Gateway config
 		"X-Twingate-Groups":             "OnCall,Engineering",
 		"X-Twingate-Client-Geo-LatLong": "37.5,-122.4",
 		"X-Twingate-Client-Geo-City":    "San Mateo",
 		"X-Twingate-Client-Geo-Region":  "CA",
 		"X-Twingate-Client-Geo-Country": "US",
+		// From GAT Token
+		"X-Twingate-Username": "alex@acme.com",
+		"X-GAT-Token-Header":  "value preserved",
```
minhtule: `TestRewrite` already covers the GAT rewrite + override behavior at the unit level. I would keep just one GAT header here to prove the end-to-end wiring and drop the duplicated override assertions. Integration tests shouldn't have to cover all cases that unit tests can.

minhtule (dismissed):
Fix & merge!

internal/webapphandler/handler.go
```diff
@@ -80,5 +80,21 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
...
+			return fmt.Errorf("header %q: %w", headerName, err)
+		}
+
+		headerValue, err := tmpl.Evaluate(variables)
+		if err != nil {
+			return fmt.Errorf("header %q: %w", headerName, err)
+		}
```
minhtule, line 99: Unlike the template in the config, we can't validate headers in the GAT so I think we should gracefully handle malformed or unsupported headers in the token: ignore them and continue. This would allow us adding new headers without breaking existing gateways.

clement0010: good call, I updated it in https://github.com/Twingate/gateway/pull/365/changes/ec3a8d95b171bc68d231e94720b4b96976e65dc1

internal/webapphandler/handler_test.go
clement0010: There's actually a test named `"GAT request header rewrites override config headers on conflict"` below. I will combine them, i think it's nicer

internal/webapphandler/handler_test.go
```diff
@@ -122,32 +122,34 @@ func TestRewrite(t *testing.T) {
...
-				"X-Config":   "Dont override",
-				"X-Username": "Overridden by GAT Token",
-			},
+			headers: map[string]string{},
 			wantHeaders: map[string]string{
-				"X-Config":   "Dont override",
-				"X-Username": "alice@acme.com",
+				"X-GAT-Good":      "alice@acme.com",
+				"X-GAT-Malformed": "",
+				"X-GAT-Unknown":   "",
```
minhtule: ~~I think we shouldn't even include the header.~~ The code doesn't seem to add these header. Why do we need to define them here?

internal/webapphandler/handler_test.go
minhtule: Oops, I didn't see it. It's actually nicer to separate 😅

internal/webapphandler/handler.go
```diff
@@ -80,5 +80,26 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
...
 	}
 
+	// Per-resource request header rewrites from the GAT are applied last, so they override
+	// any config headers with the same name. Malformed or unsupported headers are skipped
+	// rather than failing the request.
+	for headerName, value := range conn.GATClaims().Resource.GatewayMetadata.RequestHeaderRewrites {
+		tmpl, err := template.New(value)
+		if err != nil {
+			conn.Logger.Debug("skipping GAT request header rewrite", zap.String("header", headerName), zap.String("value", value), zap.Error(err))
+
```
minhtule: I think these logs should be warning.

internal/webapphandler/handler_test.go
minhtule: I also would include a header set by the config and when the GAT header is malformed/invalid, we should fallback to the config header.

internal/webapphandler/handler.go
```diff
@@ -80,5 +80,26 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
...
+			continue
+		}
+
+		headerValue, err := tmpl.Evaluate(variables)
+		if err != nil {
+			conn.Logger.Debug("skipping GAT request header rewrite", zap.String("header", headerName), zap.String("value", value), zap.Error(err))
```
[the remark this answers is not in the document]
clement0010: Updated in [88d1c08](https://github.com/Twingate/gateway/pull/365/commits/88d1c08aac282e5e98699d859d6d894ac2c933fa)

internal/webapphandler/handler_test.go
clement0010: > The code doesn't seem to add these header. Why do we need to define them here?

It's because `proxyReq.Out.Header.Get()` returns an empty string `""` for absent keys. We use it in this assertion `assert.Equal(t, wantValue, proxyReq.Out.Header.Get(name))`. I split the test to `TestRewrite_SkipsInvalidGATHeaders` so we can assert it is absent instead of an empty string

internal/webapphandler/handler.go
clement0010: Updated in [88d1c08](https://github.com/Twingate/gateway/pull/365/commits/88d1c08aac282e5e98699d859d6d894ac2c933fa)

internal/webapphandler/handler.go
```diff
@@ -80,5 +80,26 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
...
+
+			continue
+		}
+
+		headerValue, err := tmpl.Evaluate(variables)
+		if err != nil {
+			conn.Logger.Warn("skipping GAT request header rewrite", zap.String("header", headerName), zap.Error(err))
+
+			continue
+		}
```
minhtule, line 99: nit: I would create a `template.Evaluate(tmpl, variables)` to combine `template.New()` and `Template.Evaluate()` so we can simplify the error handling
```go
headerValue, err := template.New(value, variables)
if err != nil {
    conn.Logger.Warn("skipping GAT request header rewrite", zap.String("header", headerName), zap.Error(err))

    continue
}
```

minhtule (approved):
LGTM 👍

PR #371, opened 2026-07-02
fix(ssh): remove auto-generated SSH CA option

minhtule:

- Issue: #369

## Changes

- Remove the auto-generated in-memory CA; `ssh.ca` (`manual` or `vault`) is now required when SSH is enabled
- Helm chart now fails rendering when `ssh.enabled` is true and `ssh.ca` is unset (previously only when `replicaCount > 1`)

## Notes

The auto-generated CA lived only in memory, so its identity changed on every restart and
across replicas, causing host-key mismatch warnings. Requiring an explicit CA keeps it stable.


internal/sshhandler/ca.go
```diff
@@ -62,46 +62,21 @@ func (c *caConfig) Start(ctx context.Context) error {
...
+	case config.Manual != nil:
 		return newManualCA(config.Manual.PrivateKeyFile, logger)
-	}
-
-	if config.Vault != nil {
+	case config.Vault != nil:
 		return newVaultCA(config.Vault, logger)
+	default:
+		return nil, gatewayconfig.ErrMissingCAConfig
 	}
```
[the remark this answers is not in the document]
minhtule: The config validation doesn't allow both `manual` and `vault` to be set. We don't need to repeat that validation here.

deploy/gateway/values.schema.json
```diff
@@ -154,7 +154,7 @@
         },
         "ca": {
           "type": "object",
-          "description": "CA configuration. If neither manual nor vault is configured, auto-generated CA is used.",
+          "description": "CA configuration. Exactly one of manual or vault must be configured when SSH is enabled.",
```
clement0010, line 157: nit: would be nicer if we could enforce this rule at the schema-level. Something like:
```
 "if": {
        "properties": {"enabled": {"const": true}},
        "required": ["enabled"]
      },
      "then": {
        "required": ["ca"],
        "properties": {
          "ca": {
            "oneOf": [
              {"required": ["manual"]},
              {"required": ["vault"]}
            ]
          }
        }
      },
```

internal/sshhandler/ca.go
```diff
@@ -62,46 +62,21 @@ func (c *caConfig) Start(ctx context.Context) error {
...
 
 // newCAFromConfig creates CAs based on the provided configuration.
-// - If config is nil or both Manual and Vault are nil, creates an auto-generated CA.
 // - If config.Manual is set, creates an embedded CA with the provided key files.
 // - If config.Vault is set, creates Vault-backed CAs.
 func newCAFromConfig(config *gatewayconfig.SSHCAConfig, logger *zap.Logger) (*caConfig, error) {
 	if config == nil {
-		return newAutoGeneratedCA(logger)
+		return nil, gatewayconfig.ErrMissingCAConfig
 	}
```
clement0010: Do we still need this check if either `manual` or `vault` must be provided?

CLAUDE.md
clement0010:
```suggestion
- SSH server with CA-signed certificates (manual/Vault)
```

clement0010 (approved):
Fix & merge!

internal/sshhandler/ca.go
minhtule: Good catch, it was unreachable: the only caller passes `&sshCfg.CA` (a value field, never nil) and `SSHCAConfig.Validate()` already enforces that one of `manual`/`vault` is set. Removed in c6bdcfa by changing the parameter to a value type, so nil is impossible by construction.

deploy/gateway/values.schema.json
minhtule: Update in https://github.com/Twingate/gateway/pull/371/commits/fd5bb57d1afb777e59b3ca894a0f3f1d1f6a7519. Also remove the check at runtime in the `deployment.yaml` template.

deploy/gateway/tests/gateway-configmap_test.yaml
```diff
@@ -383,9 +354,19 @@ tests:
...
     asserts:
       - failedTemplate:
           errorPattern: "ssh.gateway.username is required"
+  - it: should error when SSH is enabled but ca is not specified
+    set:
+      ssh:
+        ca: null
+    asserts:
+      - failedTemplate:
+          errorPattern: "at '/ssh': missing property 'ca'"
```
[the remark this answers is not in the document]
minhtule: Omit this means `ca: {}` (the default in `values.yaml`)

sghiocel (approved):

PR #372, opened 2026-07-02
chore(ssh): document Vault CA config fields and mark inline credentials dev/test-only

minhtule:
## Changes

- Document Vault CA configuration in the Helm chart (`values.yaml`, `values.schema.json`) and `internal/config/config.go`; inline credentials are marked dev/test-only


clement0010 (approved):
LGTM 👍🏻

PR #376, opened 2026-07-03
feat: hot-reload SSH manual CA private key

clement0010:
## Related Tickets

Closes #375

## Changes

- The manual SSH CA private key file is now watched for changes, so rotating the CA key on disk takes effect without restarting the gateway.
- On rotation, the gateway's SSH host certificate is immediately re-signed with the new CA key, so clients receive updated trust material right away; user certificates are issued per connection and pick up the new key on the next connection.
- A bad rotation (unreadable or invalid key file) keeps the last good key and logs an error; startup still fails fast on an invalid key, as before.
- Each rotation logs the new CA public key so it can be distributed to upstream and client trust stores.


internal/sshhandler/key_reloader.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil
+	}
+
+	if err := watcher.Remove(event.Name); err != nil {
+		kr.logger.Info("failed to remove file watch, it may have been deleted", zap.Error(err))
+	}
+
+	err := watcher.Add(event.Name)
+
+	return err
```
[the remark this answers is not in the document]
clement0010: This is invalid, on Linux/inotify, it would emit a chmod event, so handleWatchEvent calls load(). We also have a test `TestKeyReloaderReloadWhenFileReplacedAtomically` to cover it.

internal/sshhandler/cert.go
```diff
@@ -109,21 +111,24 @@ func (s *autoRenewingCertSigner) renewalLoop(ctx context.Context) error {
...
 		select {
 		case <-timer.C:
-			nextRenewal, err := s.updateCertSigner(ctx)
-			if err != nil {
-				timer.Reset(retryInterval)
+		case <-s.renewCh:
+			s.logger.Info("CA key reloaded, re-signing certificate", zap.String("cert_type", s.certReq.certType.String()))
+		case <-ctx.Done():
+			return ctx.Err()
+		}
```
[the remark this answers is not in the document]
clement0010: Timer channels are unbuffered since `Go 1.23`, so there's never a stale value parked in `timer.C`, the `Reset` atomically discards any pending fire.

internal/sshhandler/key_reloader_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+		assert.Equal(c, newPublicKey.Marshal(), signer.PublicKey().Marshal())
+	}, time.Second, 5*time.Millisecond)
+}
+
+// TestKeyReloaderReloadWhenKubernetesSecretUpdated simulates the kubelet AtomicWriter
+// update sequence for a projected secret volume:
+//
+//	mount/ca -> ..data/ca, ..data -> ..timestamp1, update swaps ..data to ..timestamp2 and removes ..timestamp1
+func TestKeyReloaderReloadWhenKubernetesSecretUpdated(t *testing.T) {
```
clement0010, line 70: Based on the atomic write algorithm described [here](https://pkg.go.dev/github.com/kubernetes-sigs/Controller-runtime@v0.1.12/pkg/webhook/internal/cert/writer/atomic#AtomicWriter.Write)

internal/sshhandler/key_reloader.go
```diff
@@ -0,0 +1,132 @@
...
+		select {
+		case kr.reloadCh <- struct{}{}:
+		default:
+		}
+	}
+
+	return nil
+}
+
+func (kr *keyReloader) watch(ctx context.Context) error {
```
clement0010, line 79: This logic is similar to the watch function in [`cert_reloader.go`](https://github.com/Twingate/gateway/blob/6bf74c45e7fcabc44b2afe6603ce53f83ff730f8/internal/connect/cert_reloader.go#L64). I'm keeping them separated for now, if there's a third reloader, we could create a common `reloader` package and reuse them

internal/sshhandler/config.go
```diff
@@ -140,7 +140,12 @@ func (c *Config) GetDownstreamConfig(ctx context.Context) (*ssh.ServerConfig, er
 		ttl:       c.hostCertTTL,
 	}
 
-	hostCertSigner, err := newAutoRenewingCertSigner(ctx, c.caConfig.GatewayHostCA, hostCertRequest, c.hostSigner, c.logger)
+	var renewCh <-chan struct{}
+	if c.caConfig.keyReloader != nil {
+		renewCh = c.caConfig.keyReloader.reloadCh
+	}
+
+	hostCertSigner, err := newAutoRenewingCertSigner(ctx, c.caConfig.GatewayHostCA, hostCertRequest, c.hostSigner, renewCh, c.logger)
```
minhtule: This `renewCh` is passed from the `keyReloader` to the `caConfig` to the `newAutoRenewingCertSigner`, and create unnecessary coupling. Conceptually, the CA can inform when it rotates its key. We could do something like
```go
// in ca.go

// rotatingCA is a CA whose signing key can rotate during the process lifetime.
// The channel returned by rotated receives a value after each rotation.
type rotatingCA interface {
	rotated() <-chan struct{}
}

// embeddedCA can rotate it key so it implements the rotatingCA interface
type embeddedCA struct {
	getSigner func() ssh.Signer
	rotateCh  <-chan struct{}
}

func (ca *embeddedCA) rotated() <-chan struct{} {
	return ca.rotateCh
}

// in cert.go
func newAutoRenewingCertSigner(ctx context.Context, ca ca, certReq *certificateRequest, keySigner ssh.Signer, logger *zap.Logger) (*autoRenewingCertSigner, error) {
	certSigner := &autoRenewingCertSigner{ca: ca, certReq: certReq, keySigner: keySigner, logger: logger}

	// When the CA can rotate its key, re-sign as soon as it does rather than
	// waiting for the certificate to near expiry.
	if rc, ok := ca.(rotatingCA); ok {
		certSigner.renewCh = rc.rotated()
	}

	// ...
	return certSigner
}
```

internal/sshhandler/key_reloader.go
minhtule: I think we should create a generic watcher for a list of files. We'll need it to hot reload the config file. But let's do it in a follow up PR.

internal/sshhandler/ca.go
```diff
@@ -43,12 +43,17 @@ type caConfig struct {
 	GatewayUserCA  ca // Signs Gateway's user certificates (presented to upstreams)
 	UpstreamHostCA ca // Verifies upstream host certificates (only publicKey is used). If nil, defaults to TOFU verification with upstream's public key.
 
-	vault *Vault // Vault client for token lifecycle management (nil for non-Vault CAs)
+	vault       *Vault       // Vault client for token lifecycle management (nil for Manual CAs)
+	keyReloader *keyReloader // Reloads the CA private key on file change (nil for Vault CA)
```
minhtule, line 53: This struct is holding data for different CA providers. It would become unmaintainable as we support more CAs. We could create an interface and then specific struct for manual and Vault providers. For example,

```go
type caProvider interface {
	// Start runs background CA maintenance until the context is canceled.
	Start(ctx context.Context) error

	// gatewayHostCA signs the Gateway's host certificates (presented to clients).
	gatewayHostCA() ca
	// gatewayUserCA signs the Gateway's user certificates (presented to upstreams).
	gatewayUserCA() ca
	// upstreamHostCA verifies upstream host certificates (only its publicKey is used).
	// A nil result selects TOFU verification with the upstream's public key.
	upstreamHostCA() ca
}

type manualCAProvider struct {
	ca *manualCA
}

type vaultCAProvider struct {
	vault *Vault

	host     *vaultCA
	user     *vaultCA
	upstream *vaultCA
}
```

I would do this in a separate follow-up PR.

internal/sshhandler/key_reloader_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+	requireKeyReloaderSigner(t, keyReloader, oldPublicKey)
+
+	newKeyPEM, newPublicKey := generateCAKey(t)
+	newDataDir := filepath.Join(mount, "..timestamp2")
+	require.NoError(t, os.Mkdir(newDataDir, 0750))
+	require.NoError(t, os.WriteFile(filepath.Join(newDataDir, "ca"), newKeyPEM, 0600))
+	require.NoError(t, os.Symlink("..timestamp2", filepath.Join(mount, "..data_tmp")))
+	require.NoError(t, os.Rename(filepath.Join(mount, "..data_tmp"), filepath.Join(mount, "..data")))
+	require.NoError(t, os.RemoveAll(oldDataDir))
```
minhtule, line 93: I would add a comment here because the trigger is subtle. Swapping the `..data` symlink emits no event; it's removing the old data dir that makes the watched path fire (Chmod on Linux, Create on macOS), at which point `..data` already resolves to the new key so `load()` picks it up. The `Remove` event just re-adds the watch.

internal/sshhandler/key_reloader_test.go
minhtule: This is another reason we should create a common file watcher for TLS cert and SSH key files

internal/sshhandler/key_reloader_test.go
clement0010: Added in [30e2e32](https://github.com/Twingate/gateway/pull/376/commits/30e2e32314cce4af62dac00d9cb09d69d7f9333f)

internal/sshhandler/config.go
clement0010: Updated in [ba72199](https://github.com/Twingate/gateway/pull/376/commits/ba72199d92235daebc6aa321f2da5a771fa1f843)

internal/sshhandler/ca.go
```diff
@@ -37,18 +37,29 @@ type ca interface {
 	sign(ctx context.Context, req *certificateRequest) (*ssh.Certificate, error)
 }
 
+// rotatingCA is a CA whose signing key can rotate during the process lifetime.
+// The channel returned by rotated receives a value after each rotation.
+type rotatingCA interface {
+	rotated() <-chan struct{}
```
minhtule, line 43: I think we should note that this signal is only for 1 consumer, which is good enough for now.

minhtule (approved):
LGTM 👍

sghiocel (approved):

PR #380, opened 2026-07-06
test: split SSH key config parsing and key generation tests

minhtule:

- Issue: N/A

## Changes

- Split `TestNewKeyConfigAndGenerate_ByKeyType` into two tests:
  - `TestNewKeyConfig`: all 12 parsing cases, no key generation
  - `TestKeyConfigGenerate`: one case per `Generate` code path (ed25519, ECDSA 256/384/521, RSA 2048)

## Notes

- Every parsing case previously ran `Generate`, so the RSA cases dominated the test runtime (1.5-4s out of ~12s total, [sample build](https://github.com/Twingate/gateway/actions/runs/28804917304/job/85417827252))
- RSA-4096 was generated twice: once via the `rsa` key type with default bits, once via `ssh-rsa`
- The RSA branch in `Generate` does not vary by key size, so a single 2048-bit case covers it
- Test runtime drops to ~0.2s ([sample build](https://github.com/Twingate/gateway/actions/runs/28810211844/job/85435919214?pr=380)) with no loss of branch coverage

### DIFF
--- a/internal/sshhandler/keys_test.go
+++ b/internal/sshhandler/keys_test.go
@@ -12,7 +12,7 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
-func TestNewKeyConfig_ByKeyType(t *testing.T) {
+func TestNewKeyConfig(t *testing.T) {
 	tests := []struct {
 		name       string
 		keyType    string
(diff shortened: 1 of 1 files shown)


clement0010 (approved):
LGTM 👍🏻

PR #381, opened 2026-07-08
chore: Rewrite SSH unit tests

minhtule:
## Changes

Rewrite the `internal/sshhandler` tests against real SSH connections and remove the test-only protocol seams they existed to serve. Behavior-preserving.

- Remove the SSH-protocol test seams (interfaces, factories, `mocks.go`) so production wires concrete types directly.
- Rewrite the tests against real SSH connections over loopback TCP, organized by observable SSH behavior instead of by source function.
- Minor production edge cases surfaced by dropping the seams (unhandled or now-unused code); called out inline on the diff.

## Notes

The suite is organized by SSH layer, one test file per layer, so each layer's tests exercise the real stack below it:

- **proxy** (`proxy_test.go`, covers `proxy.go`): transport handshake, user auth (host cert + user cert), and wiring an accepted connection through to its upstream.
- **connection** (`conn_pair_test.go`, covers `conn_pair.go`): connection proxying, i.e. channel forwarding and global-request forwarding between the two sides.
- **channel** (`channel_pair_test.go`, covers `channel_pair.go` + `request_handler.go` + `channel_copier.go`): channel proxying, split by channel type:
  - non-session channels: channel data and channel-request forwarding, flowing immediately in both directions.
  - session channels: data is gated until a session-start request (shell/exec/subsystem) is forwarded, and shell sessions are recorded.

Each test drives its layer over real SSH connections and asserts on what the far end observes, not on internal calls, so the internals can be refactored freely without touching tests. The only test double is the session recorder.


internal/sshhandler/channel_copier.go
```diff
@@ -33,6 +35,12 @@ type ChannelCopyPair struct {
 
 	// Tap for recording a channel's data from Src channel
 	Tap io.Writer
+
+	// eofTimeout bounds the wait for a pending-request flush or a Src close before sending EOF.
+	eofTimeout time.Duration
+
+	// closeTimeout bounds the wait for the peer to close the channel before force-closing it.
+	closeTimeout time.Duration
```
minhtule, line 43: Add these timeout fields so they can be lowered in unit tests without overriding the package-level defaults, which would race across parallel tests.

internal/sshhandler/channel_pair.go
```diff
@@ -49,36 +52,37 @@ type SSHChannelPair struct {
...
 
 	// Factory for creating session recorders
 	recorderFactory SessionRecorderFactory
 
+	// sessionStartTimeout bounds the wait for a session-start request before serve() gives up.
+	sessionStartTimeout time.Duration
+
+	// channelEOFTimeout and channelCloseTimeout configure the data copiers' teardown waits.
+	channelEOFTimeout   time.Duration
+	channelCloseTimeout time.Duration
```
minhtule, line 69: Add these timeout fields so they can be lowered in unit tests without overriding the package-level defaults, which would race across parallel tests.

internal/sshhandler/proxy.go
```diff
@@ -129,11 +70,6 @@ func (p *SSHProxy) Start(ctx context.Context, listener net.Listener) error {
 			break
 		}
 
-		// No longer serving
-		if conn == nil {
-			break
-		}
```
minhtule, line 135: This was used only in unit test. When `err` is nil, there must be a `conn`.

internal/sshhandler/request_handler.go
```diff
@@ -268,24 +218,13 @@ func (h *SSHRequestHandler) handleRequests() SSHSessionSignals {
...
 
 		return requestErr
 	}
 
-	// Reply to the original request with the reply from forwarded request
-	if request.GetWantReply() {
-		if replyErr := request.Reply(reply, nil); replyErr != nil {
-			return replyErr
-		}
-	}
```
minhtule, line 288: `request.GetWantReply()` check (equivalent to `request.WantReply ==  true` check) is unnecessary because `request.Reply()` does nothing when `request.WantReply ==  false` ([code](https://github.com/golang/crypto/blob/fdfe1f8531a1adcc300c8eba98cb372044826d62/ssh/channel.go#L97-L99)). This behavior is documented publicly [here](https://pkg.go.dev/golang.org/x/crypto/ssh#Request.Reply)
> Reply sends a response to a request. It must be called for all requests where WantReply is true and is a no-op otherwise.

internal/sshhandler/request_handler.go
```diff
@@ -268,24 +218,13 @@ func (h *SSHRequestHandler) handleRequests() SSHSessionSignals {
...
-
-	return nil
+	return request.Reply(reply, nil)
 }
```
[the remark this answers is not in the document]
minhtule: See https://github.com/Twingate/gateway/pull/381#discussion_r3548113009

internal/sshhandler/conn_pair.go
```diff
@@ -240,20 +202,16 @@ func (c *SSHConnPair) forwardGlobalRequests(requests <-chan *ssh.Request, dst ss
...
 		c.logger.Error("Failed to close upstream connection", zap.Error(err))
 	}
 }
 
 func replyToGlobalRequest(req *ssh.Request, ok bool, payload []byte, logger *zap.Logger) {
-	if !req.WantReply {
-		return
-	}
-
 	if err := req.Reply(ok, payload); err != nil {
```
minhtule, line 257: The check is unnecessary similar to channel request in https://github.com/Twingate/gateway/pull/381#discussion_r3548113009

internal/sshhandler/channel_pair.go
```diff
@@ -143,8 +147,8 @@ func (c *SSHChannelPair) serve() {
 		logger:            c.logger,
 		sshChannelCtx:     targetChannelCtx,
 		flushTrigger:      targetEOFTrigger,
-		sourceRequestChan: c.targetChannelRequests,
-		targetChannel:     c.sourceChannel,
+		sourceRequestChan: c.target.requests,
+		targetChannel:     c.source.ch,
 		onPtyRequest:      nil,
 		onWindowChange:    nil,
 	}
```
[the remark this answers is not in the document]
minhtule: yes, known issue. Will fix separately!

sghiocel (approved):

internal/sshhandler/channel_pair_test.go
```diff
@@ -5,365 +5,666 @@ package sshhandler
...
+			require.NoError(t, other.ch.Close())
+			assertEOF(t, initiator.ch)
+			waitReqChanClosed(t, initiator.requests)
+			waitDone(t, done)
+		})
+	}
+}
 
-	return args.Error(0)
+func TestChannelPair_TeardownTimeout(t *testing.T) {
```
clement0010, line 212: I think we can also add a test `TestChannelPair_EOFFlushTimeout` to cover [this case](https://github.com/Twingate/gateway/pull/381/changes#diff-f9a38ff02468eece1dfafcdc1fddf349f94040347a6be21f33e5ac9cec986bd4R70-R81)

clement0010 (approved):
LGTM 👍🏻

internal/sshhandler/channel_pair_test.go
minhtule: Added in https://github.com/Twingate/gateway/commit/71fe3c0a61764ec0d8c5c6f75f880b5717d28718

PR #382, opened 2026-07-08
feat: Support resource alias translation in CONNECT message

clement0010:
Branched out from https://github.com/Twingate/gateway/pull/363

## Changes

Gateway should support translating the alias address in the CONNECT message to the upstream address alongside the existing downstream→upstream port mapping.

It should also validate that the CONNECT address must match the token's `resource.address` (exact or wildcard) or its `resource.alias` (exact match).

A GAT carrying an alias for a wildcard address is rejected at token validation, so an alias always maps to a concrete upstream address.


internal/connect/connect.go
```diff
@@ -171,11 +160,24 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+	case matchResourceAddress(resource.Address, host):
+	case matchResourceAlias(resource.Alias, host):
+		host = resource.Address
+	default:
+		return "", &HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("CONNECT destination address %s does not match token resource address %s", host, resource.Address),
+			Err:     nil,
+		}
+	}
```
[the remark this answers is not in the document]
clement0010: The GAT token `alias` should be empty when the `address` is a wildcard pattern.

internal/token/gat_claims.go
```diff
@@ -68,7 +70,15 @@ func (p GATClaims) Validate() error {
...
 	}
 
-	return validatePort(p.Resource.GatewayMetadata.Upstream.Port, "resource.gateway_metadata.upstream.port")
+	if err := validatePort(p.Resource.GatewayMetadata.Upstream.Port, "resource.gateway_metadata.upstream.port"); err != nil {
+		return err
+	}
+
+	if p.Resource.Alias != "" && IsWildcardAddress(p.Resource.Address) {
+		return fmt.Errorf("%w: %w: address %q, alias %q", jwt.ErrTokenInvalidClaims, errWildcardAlias, p.Resource.Address, p.Resource.Alias)
+	}
```
clement0010: While the GAT token could never return a wildcard address with an alias, I added this to ensure we could never rewrite the upstream address to a wildcard.

internal/connect/connect.go
```diff
@@ -171,11 +160,34 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
...
+// resolveUpstreamAddress verifies the CONNECT destination against the GAT
+// resource and maps it to the upstream address for backend forwarding: a host
+// matching the resource alias maps to the resource address, and the downstream
+// port maps to the upstream port. Both ports are validated for presence and
+// range when the token is parsed.
+func resolveUpstreamAddress(host, port string, resource token.Resource) (string, *HTTPError) {
+	switch {
+	case matchResourceAddress(resource.Address, host):
+		//	Keep existing upstream address
+	case matchResourceAlias(resource.Alias(), host):
```
minhtule: Why do we only match again the first alias? The GAT returns a list so we should match again any element in that list.

internal/connect/connect.go
```diff
@@ -144,18 +144,7 @@ func (v *MessageValidator) ParseConnect(req *http.Request, ekm []byte) (connectI
 			}
 	}
```
minhtule: I would move these logic into the `resolveUpstreamAddress` too

minhtule (changes requested):
Now that we need to validate and translate both hostname and port of the address, I would fold this into https://github.com/Twingate/gateway/pull/363 to have a coherent change set to review. If I need to split it, I would split the GAT changes into a small PR that can be reviewed and merged immediately.

internal/connect/connect.go
clement0010: Yeah you're right it should match to any elements in the list. Updated in [e03d311](https://github.com/Twingate/gateway/pull/382/commits/e03d311a1078cec46ca1905383bd0b78629fd640)

clement0010:
> Now that we need to validate and translate both hostname and port of the address, I would fold this into #363 to have a coherent change set to review. If I need to split it, I would split the GAT changes into a small PR that can be reviewed and merged immediately.

@minhtule I merged this PR to #363, it probably needs your re-review there

PR #398, opened 2026-07-13
fix(ssh): recover panics in connection-serving goroutines

minhtule:

Issue: https://github.com/Twingate/gateway/issues/396

## Changes

- Recover and log (with stack trace) panics in every per-connection goroutine: connection serving, global-request and channel forwarding, channel copying, and request handling. A panic now tears down only its own connection instead of crashing the process and every active session.


internal/sshhandler/request_handler.go
```diff
@@ -172,7 +172,10 @@ func (h *SSHRequestHandler) handleRequests() SSHSessionSignals {
 	}
 
 	go func() {
+		// Recover panics so a failure handling one request cannot crash the whole process;
+		// closing finished still lets the channel pair tear down.
 		defer close(sessionSignals.finished)
+		defer recoverPanic(h.logger)
```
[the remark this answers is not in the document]
minhtule: Reordered in ae74c59: recoverPanic is now the outermost defer so it also covers panics from deferred cleanup.

internal/sshhandler/proxy.go
```diff
@@ -72,6 +72,9 @@ func (p *SSHProxy) Start(ctx context.Context, listener net.Listener) error {
 
 		// Serve SSH connection in a separate goroutine
 		go func() {
+			// A panic while serving one connection must not bring down the whole proxy.
+			defer recoverPanic(p.config.logger)
+
 			err := p.serveConn(ctx, conn.(connect.Conn))
```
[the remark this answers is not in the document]
minhtule: There is no capture race: `conn` is declared with `:=` inside the loop body, so each iteration gets a fresh variable (true even before Go 1.22 per-iteration semantics; only `for`-clause variables were shared).

The close-on-panic point is valid though; added `defer conn.Close()` in ae74c59 so a recovered panic cannot leave the accepted socket open (matches what net/http's `conn.serve` does).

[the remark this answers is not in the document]
minhtule: The loop-variable race claim is wrong; conn is declared with := inside the loop body, so each iteration gets a fresh variable.

internal/sshhandler/proxy.go
```diff
@@ -178,26 +181,33 @@ func (p *SSHProxy) serveConn(ctx context.Context, conn connect.Conn) error {
...
 
 	logger.Info("SSH connection closed", zap.Any("ssh", sshCtx.withConnectionClose(sshConnPair.ChannelsOpened())))
 
-	// Remove the closed SSH connection pair from the map
-	p.mu.Lock()
-	delete(p.connsMap, sshConnPair)
-	p.mu.Unlock()
-
-	p.wg.Done()
-
```
minhtule, line 197: Use `defer` to make sure the proxy can shutdown when a connection panics.

internal/sshhandler/proxy.go
```diff
@@ -178,26 +181,33 @@ func (p *SSHProxy) serveConn(ctx context.Context, conn connect.Conn) error {
...
 	return nil
 }
 
+// recoverPanic is deferred in every connection-serving goroutine to keep a panic in one
+// connection from crashing the whole process.
+func recoverPanic(logger *zap.Logger) {
+	if recovered := recover(); recovered != nil { //nolint:revive // recoverPanic itself is the deferred function
+		logger.Error("Recovered from panic", zap.Any("panic", recovered), zap.Stack("stacktrace"))
+	}
+}
```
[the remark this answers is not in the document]
minhtule: Not fixed. We'll make sure to always pass a logger.

internal/sshhandler/request_handler.go
```diff
@@ -172,6 +172,7 @@ func (h *SSHRequestHandler) handleRequests() SSHSessionSignals {
 	}
 
 	go func() {
+		defer closeOnPanic(h.logger, func() { _ = h.targetChannel.Close() })
```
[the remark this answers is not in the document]
minhtule: This is addressed in https://github.com/Twingate/gateway/pull/397

internal/sshhandler/channel_pair.go
```diff
@@ -218,3 +218,8 @@ func (c *SSHChannelPair) serve() {
 
 	copier.start()
 }
+
+func (c *SSHChannelPair) close() {
```
clement0010, line 216: I think we are missing a test for this?

clement0010 (approved):
One comment. LGTM 👍🏻

internal/sshhandler/channel_pair.go
minhtule: Added in https://github.com/Twingate/gateway/pull/398/commits/8181cb12af9c978782c63f93030d3b45c436e27a

internal/sshhandler/proxy.go
sghiocel, line 152: Do we need to add a panic handler for this connection (similar to what we do for the downstream), in case we panic prior to calling `sshConnPair.serve()` (which has its own panic handler) ?

internal/sshhandler/proxy.go
```diff
@@ -178,26 +180,32 @@ func (p *SSHProxy) serveConn(ctx context.Context, conn connect.Conn) error {
...
-
-	p.wg.Done()
-
 	return nil
 }
 
+func closeOnPanic(logger *zap.Logger, closeFn func()) {
+	if recovered := recover(); recovered != nil { //nolint:revive // closeOnPanic itself is the deferred function
+		logger.Error("Recovered from panic", zap.Any("panic", recovered), zap.Stack("stacktrace"))
+		closeFn()
```
sghiocel: Maybe we also want to attempt a recovery if this `closeFn()` panics, and log if this happened?

minhtule: Added in https://github.com/Twingate/gateway/pull/398/commits/8130ac62e4215537b9755a290afbe109b20be0c9

internal/sshhandler/proxy.go
minhtule: Yes, good call! Added in https://github.com/Twingate/gateway/pull/398/commits/3772437ab247f8542ca42927fb5f7503789f27d3

sghiocel (approved):

PR #404, opened 2026-07-19
feat(ssh): log channel request outcomes in audit logs

minhtule:

- Issue #396

## Changes

- Log one audit line per forwarded channel request, carrying the target's reply (`accepted`) when the request asked for one; previously the line was emitted before forwarding, so the outcome was never recorded.
- Log at Info by default, and at Debug for the request types with no audit value (`pty-req`, `window-change`, `signal`, `xon-xoff`, `break`, `exit-status`, `exit-signal`), so an unrecognized request type stays visible at Info.
  + Compared with the current behavior, `pty-req` moves from Info to Debug and unrecognized types move from Debug to Info.
- Reply exactly once per request, rejecting on the paths that do not forward (duplicate session start, forward error).

## Notes

- Mirrors the global-request audit logging from #390 (log every request and its outcome, including denied attempts) at the channel-request level.

### DIFF
--- a/internal/sshhandler/conn_pair.go
+++ b/internal/sshhandler/conn_pair.go
@@ -265,8 +265,7 @@ func (c *SSHConnPair) handleGlobalRequest(req *ssh.Request, dst ssh.Conn, disall
 	}
 }
 
-// forwardGlobalRequest forwards a global request to dst and returns the reply to send back. The
-// named returns default to a rejection, so every early return rejects the request.
+// forwardGlobalRequest forwards a global request to dst and returns the reply to send back.
 func (c *SSHConnPair) forwardGlobalRequest(req *ssh.Request, dst ssh.Conn, disallowedTypes map[string]bool, source, target string) (accepted bool, replyPayload []byte) {
 	extra, parseErr := globalRequestLogFields(req.Type, req.Payload)
 
--- a/internal/sshhandler/conn_pair.go
+++ b/internal/sshhandler/conn_pair.go
@@ -156,13 +156,13 @@ func (c *SSHConnPair) serve() {
 	c.wg.Go(func() {
 		defer closeOnPanic(c.logger, c.close)
 
-		c.forwardGlobalRequests(c.downstream.requests, c.upstream.conn, disallowedDownstreamGlobalRequests, labelDownstream, labelUpstream)
+		c.handleGlobalRequests(c.downstream.requests, c.upstream.conn, disallowedDownstreamGlobalRequests, labelDownstream, labelUpstream)
 	})
 
 	c.wg.Go(func() {
 		defer closeOnPanic(c.logger, c.close)
 
-		c.forwardGlobalRequests(c.upstream.requests, c.downstream.conn, disallowedUpstreamGlobalRequests, labelUpstream, labelDownstream)
+		c.handleGlobalRequests(c.upstream.requests, c.downstream.conn, disallowedUpstreamGlobalRequests, labelUpstream, labelDownstream)
 	})
 
 	// Forward channels in both directions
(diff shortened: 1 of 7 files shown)


internal/sshhandler/request_handler.go
```diff
@@ -86,22 +119,16 @@ func (h *SSHRequestHandler) handleRequest(req *ssh.Request, sessionSignals SSHSe
 		if h.onPtyRequest != nil {
 			h.onPtyRequest(ptyReq)
 		}
-
-		shouldLog = true
```
minhtule, line 90: This PR changes the log level for `pty-req` channel request from `INFO` to `DEBUG`. The dimension of the terminal is part of the asciicast header so I think lower this log level would reduce the noise.

internal/sshhandler/conn_pair.go
```diff
@@ -265,34 +275,21 @@ func (c *SSHConnPair) forwardGlobalRequest(req *ssh.Request, dst ssh.Conn, disal
...
-		replyPayload []byte
-		err          error
-	)
-
-	// Reply exactly once, on every path; early returns leave the defaults, which reject the request.
-	defer func() {
-		if err := req.Reply(accepted, replyPayload); err != nil {
-			logger().Error("Failed to reply to global request", zap.Error(err))
-		}
-	}()
```
minhtule, line 279: Replace this usage of `defer` with having a helper method, which I think is cleaner and is used for channel request. So this change make both global and channel request handling similar.

internal/sshhandler/conn_pair.go
```diff
@@ -250,13 +250,23 @@ func (c *SSHConnPair) forwardChannels(channels <-chan ssh.NewChannel, targetConn
 	}
 }
 
-func (c *SSHConnPair) forwardGlobalRequests(requests <-chan *ssh.Request, dst ssh.Conn, disallowedTypes map[string]bool, source, target string) {
+func (c *SSHConnPair) handleGlobalRequests(requests <-chan *ssh.Request, dst ssh.Conn, disallowedTypes map[string]bool, source, target string) {
```
minhtule, line 253: Rename this so the helper method can be `forwardGlobalRequest`.

test/integration/testutil/ssh_client.go
```diff
@@ -53,6 +53,10 @@ func (s *SSH) baseOptions() []string {
 		s.hostname,
 		"-l", s.username,
 		"-p", s.port,
+		// Ignore the system and user ssh_config so the requests the client sends, and therefore the
+		// audit logs the gateway emits, do not vary with the machine running the tests. In
+		// particular `SendEnv LANG LC_*` would add an env request whenever LANG is set.
+		"-F", "none",
```
minhtule, line 59: This option is needed because we log the `env` request now. Different test run time might have different user's envvar that OpenSSH might send.

internal/sshhandler/request_handler.go
```diff
@@ -60,24 +72,45 @@ type windowChangeReq struct {
...
-		h.logger.Error("Failed to parse "+req.Type+" request",
+		h.logger.Error("Failed to parse channel request",
 			zap.Any("ssh", h.sshChannelCtx.withRequest(req.Type, nil)),
 			zap.Error(err))
 	}
 }
 
-// handleRequest processes and forwards a single SSH request, returning session info if applicable.
 func (h *SSHRequestHandler) handleRequest(req *ssh.Request, sessionSignals SSHSessionSignals) {
-	h.logger.Debug("Channel request", zap.Any("ssh", h.sshChannelCtx.withRequest(req.Type, nil)))
```
minhtule, line 71: Remove this so we don't log a channel request twice. For request types that are unnecessary to be audited, we'll lower the log level to debug.

internal/sshhandler/request_handler.go
```diff
@@ -60,24 +72,45 @@ type windowChangeReq struct {
 // parseRequestPayload unmarshals request payload and logs error if parsing fails.
 func (h *SSHRequestHandler) parseRequestPayload(req *ssh.Request, target any) {
 	if err := ssh.Unmarshal(req.Payload, target); err != nil {
-		h.logger.Error("Failed to parse "+req.Type+" request",
+		h.logger.Error("Failed to parse channel request",
```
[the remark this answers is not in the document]
minhtule: Failed to parse request would still be forwarded as the current behavior.

sghiocel (approved):

clement0010 (approved):
LGTM 👍🏻

PR #407, opened 2026-07-21
refactor: extract generic file reloader from cert and key reloaders

clement0010:

- Closes: https://github.com/Twingate/gateway/issues/375
- Follow-up for review comment: https://github.com/Twingate/gateway/pull/376#discussion_r3609543341

## Changes

- Add `internal/reloader`: a generic reloader that watches a list of files and invokes a caller-supplied reload callback on change, owning the fsnotify watch loop, the retry-on-failure behavior, and the Remove/Rename re-add handling for atomic renames and Kubernetes projected-volume updates.
- Migrate `sshhandler.keyReloader` and `connect.CertReloader` onto it; each now holds only its typed state (SSH signer / TLS certificate) and a `load()` callback.
- Consolidate the watch/reload behavior tests into `internal/reloader`, leaving each reloader's tests focused on its own loading semantics (signer swap and rotation signal; cert/key pairing).

## Notes
The two existing reloaders previously carried near-identical watch loops; that duplication is now removed.


internal/sshhandler/key_reloader_test.go
```diff
@@ -139,139 +90,6 @@ func TestKeyReloaderSignalsReloadOnKeyChange(t *testing.T) {
 	requireNoReloadSignal()
 }
 
-func TestKeyReloaderRecoversAfterInvalidKey(t *testing.T) {
```
clement0010, line 142: Moved all these tests to `reloader_test.go`

internal/sshhandler/key_reloader_test.go
```diff
@@ -4,100 +4,51 @@
...
-	require.NoError(t, os.Symlink("..timestamp1", filepath.Join(mount, "..data")))
-	require.NoError(t, os.Symlink(filepath.Join("..data", "ca"), filepath.Join(mount, "ca")))
+	tests := []struct {
+		name    string
+		keyFile string
+		wantErr bool
+	}{
+		{name: "valid key", keyFile: keyFile},
+		{name: "invalid key", keyFile: invalidKeyFile, wantErr: true},
+		{name: "missing file", keyFile: "nonexistent", wantErr: true},
```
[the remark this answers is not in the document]
clement0010: I think it is fine to leave as is

internal/reloader/reloader.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+}
+
+func (r *Reloader) handleWatchEvent(event fsnotify.Event, watcher *fsnotify.Watcher) error {
+	r.logger.Debug("Received watch event", zap.Any("event", event))
+
+	if !event.Has(fsnotify.Remove) && !event.Has(fsnotify.Rename) {
+		if err := r.load(); err != nil {
+			r.logger.Error("Failed to load "+r.name, zap.Error(err))
```
minhtule: I think we don't need to the `name` field that is only used in logs. Here we could use `event.Name`. In other places, we could concat the file names.

internal/reloader/reloader.go
```diff
@@ -0,0 +1,96 @@
...
+// changes, retrying the watch loop after transient failures.
+type Reloader struct {
+	name   string
+	files  []string
+	load   func() error
+	logger *zap.Logger
+}
+
+// New creates a Reloader that watches files and calls load on every change.
+func New(name string, logger *zap.Logger, load func() error, files ...string) *Reloader {
```
minhtule: Important arguments should appear first. I would change this signature to
```go
func New(files []string, load func(), logger *zap.Logger) *Reloader
```

internal/reloader/reloader_test.go
```diff
@@ -0,0 +1,295 @@
...
+	require.NoError(t, os.WriteFile(fileB, []byte("b2"), 0600))
+
+	// Initial load plus one event per changed file.
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		mu.Lock()
+		defer mu.Unlock()
+
+		assert.GreaterOrEqual(c, loads, 3)
+	}, time.Second, 5*time.Millisecond)
+}
```
minhtule, line 183: This test would pass if we only watch the 1st file e.g. changing line 52 in `reloader.go` from
```go
for _, file := range r.files {
```
to
```go
for _, file := range r.files[:1] {
```

This is because `os.WriteFile` would trigger 2 events: chmod and write. I think it's more robust to assert that we are able to get the new content instead of counting the number of watch events.

internal/connect/cert_reloader.go
```diff
@@ -22,22 +20,23 @@ type CertReloader struct {
...
+	return cr
 }
 
 func (cr *CertReloader) Run(ctx context.Context) {
-	go wait.Until(func() {
-		if err := cr.watch(ctx); err != nil {
-			cr.logger.Error("failed to watch cert and key file, will retry later", zap.Error(err))
-		}
-	}, time.Minute, ctx.Done())
+	cr.reloader.Run(ctx)
```
minhtule, line 39: nit: We might want to have a test for this to make sure the watch loop is called.

internal/sshhandler/key_reloader.go
```diff
@@ -26,22 +24,23 @@ type keyReloader struct {
...
+	return kr
 }
 
 func (kr *keyReloader) Run(ctx context.Context) {
-	go wait.Until(func() {
-		if err := kr.watch(ctx); err != nil {
-			kr.logger.Error("failed to watch CA private key file, will retry later", zap.Error(err))
-		}
-	}, time.Minute, ctx.Done())
+	kr.reloader.Run(ctx)
```
minhtule, line 43: Ditto

internal/connect/cert_reloader.go
clement0010: Added in [1b8eddf](https://github.com/Twingate/gateway/pull/407/commits/1b8eddfc2636af912fc9ba5398009c315d8fabcd)

internal/reloader/reloader_test.go
clement0010: Good idea, I updated the test in [2c04bed](https://github.com/Twingate/gateway/pull/407/commits/2c04beda0ccc2d0afeef9d74b4cc8b66c017faba)

minhtule (approved):
LGTM 👍

sghiocel (approved):

PR #408, opened 2026-07-21
refactor(ssh): replace caConfig struct with a caProvider interface

clement0010:
## Related Tickets & Documents

- Closes: https://github.com/Twingate/gateway/issues/375
- Follow-up for review comment: https://github.com/Twingate/gateway/pull/376#discussion_r3609571905

## Changes

- Replace the `caConfig` struct, which mixed the exposed CAs with backend-specific fields (`vault`, `keyReloader`), with a `caProvider` interface implemented by `manualCAProvider` and `vaultCAProvider`.
- Each provider owns its own `Start`, removing the nil-field branching in the old shared `Start`. `newCAFromConfig` returns the interface.
- Turn `upstreamHostKeyCallback` into a package function taking the upstream CA (nil selects TOFU), since it was the only reader of the old `UpstreamHostCA` field.

PR #410, opened 2026-07-21
fix(ssh): persist TOFU upstream host key pin across connections

minhtule:

- Issue: #409
- Stacked on #408 (`refactor/ct/ca-provider-interface`); base this PR on that branch.

## Changes

- Keep TOFU host-key pins in a per-address map on `manualCAProvider`, so the first key seen for an upstream is enforced across connections for the process lifetime.
- Make `upstreamHostKeyCallback` a `caProvider` method: manual mode pins via TOFU, Vault mode verifies the upstream host certificate against the upstream host CA public key.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -46,6 +46,8 @@ linters:
         - stdlib
         # custom values
         - Recorder
+        - ca
+        - caProvider
         # golang.org/x/crypto/ssh
         - Signer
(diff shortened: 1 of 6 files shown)


sghiocel (approved):

clement0010 (approved):
LGTM 👍🏻

PR #414, opened 2026-07-23
feat: support upstream TLS for Web App resources

### LINKED ISSUE #413: Support TLS upstream for Web App resource
## Summary

The Gateway currently proxies Web App traffic to the upstream over plain HTTP only. The Gateway should connect to the web app over TLS when the resource's GAT requires it, verifying the upstream certificate against operator-configured CAs and system root CAs.

## Changes

- Parse the `tls` field in the GAT token's `gateway_metadata.upstream`; tokens without the field keep plain-HTTP behavior (backward compatible).
- Add a top-level `cas` config list (`{name, certFile}`) declaring the CAs trusted for upstream TLS verification; the Gateway checks the upstream certificate against all configured CAs to find a match.
- When `upstream.tls` is true, connect over HTTPS with a TLS 1.3 minimum and verify against the configured CAs only, failing the connection when verification fails.

clement0010:

- Issue: closes #413

## Changes

- Parse the `tls` flag in the GAT token's `gateway_metadata.upstream`; tokens without the flag keep today's plain-HTTP behavior.
- Add a top-level `cas` config list (`{name, certFile}`) declaring additional CAs the Gateway trusts when verifying upstream TLS connections.
- Connect to Web App upstreams over HTTPS when the GAT enables it, verifying the upstream certificate against the system trust store plus the configured CAs with a TLS 1.3 minimum.


internal/webapphandler/config.go
```diff
@@ -35,5 +42,29 @@ func NewConfig(configRequestHeaders map[string]string, roundTripperMetrics *metr
...
+	}
+
+	return &Config{requestHeaders: headers, caPool: caPool, roundTripperMetrics: roundTripperMetrics, logger: logger}, nil
+}
+
+// newCAPool merges the configured CAs into the pool used to verify upstream TLS connections.
+// The pool never includes system roots, so an empty list rejects every upstream certificate.
+func newCAPool(cas []config.CA) (*x509.CertPool, error) {
+	pool := x509.NewCertPool()
+
```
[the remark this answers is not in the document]
clement0010: I will support hot-reload in a separate PR.

internal/webapphandler/handler.go
```diff
@@ -32,12 +34,28 @@ func NewHandler(cfg Config) *Handler {
...
 	return &Handler{proxy: proxy}
 }
 
+// createTransport clones http.DefaultTransport to preserve its proxy, timeout,
+// and HTTP/2 defaults and pins upstream TLS verification to the given CA pool.
+func createTransport(caPool *x509.CertPool) *http.Transport {
+	transport := &http.Transport{}
+	if defaultTransport, ok := http.DefaultTransport.(*http.Transport); ok {
+		transport = defaultTransport.Clone()
+	}
```
minhtule: I don't think we need to check the casting here. The `DefaultTransport` must be a `Transport` interface...
```go
transport := http.DefaultTransport.(*http.Transport).Clone()
```

internal/webapphandler/handler_test.go
```diff
@@ -240,6 +242,43 @@ func TestRewrite_PreservesClientHost(t *testing.T) {
...
 }
 
+func TestRewrite_UpstreamScheme(t *testing.T) {
+	tests := []struct {
+		name        string
+		upstreamTLS bool
+		wantScheme  string
+	}{
+		{name: "plain HTTP when tls is false", upstreamTLS: false, wantScheme: "http"},
+		{name: "HTTPS when tls is true", upstreamTLS: true, wantScheme: "https"},
```
minhtule: nit: like `HTTP`, `TLS` should be all uppercased in English context (comments and docs).

internal/webapphandler/config.go
```diff
@@ -35,5 +42,32 @@ func NewConfig(configRequestHeaders map[string]string, roundTripperMetrics *metr
...
+	for _, ca := range cas {
+		caCert, err := os.ReadFile(ca.CertFile) //nolint:gosec // The CA file is provided by the operator
+		if err != nil {
+			return nil, fmt.Errorf("ca %q: %w", ca.Name, err)
+		}
+
+		if ok := pool.AppendCertsFromPEM(caCert); !ok {
+			return nil, fmt.Errorf("ca %q: %w", ca.Name, errInvalidCACert)
+		}
+	}
```
minhtule, line 70: We don't seem to have a test for the custom CAs are being added to the CA pool. We could either unit test this or add an integration test (which also verifies the transport wiring).

minhtule (approved):
Fix & merge!

internal/webapphandler/config.go
clement0010: I think adding an integration test for HTTPS would be better since it could also include the downstream HTTPS wiring too

internal/webapphandler/config.go
```diff
@@ -35,5 +42,32 @@ func NewConfig(configRequestHeaders map[string]string, roundTripperMetrics *metr
...
+	return &Config{requestHeaders: headers, caPool: caPool, roundTripperMetrics: roundTripperMetrics, logger: logger}, nil
+}
+
+// newCAPool merges the configured CAs on top of the system cert pool to verify
+// upstream TLS connections, so publicly-signed upstreams work without configuration.
+func newCAPool(cas []config.CA) (*x509.CertPool, error) {
+	pool, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("load system cert pool: %w", err)
+	}
```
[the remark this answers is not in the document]
clement0010: The issue is outdated, will update the description

sghiocel (approved):

PR #422, opened 2026-07-24
feat: support downstream TLS for Web App resources

clement0010:

- Issue: closes #420

## Changes

- Parse the `tls` flag in the GAT token's `gateway_metadata.downstream`; tokens without the flag keep today's plaintext behavior.
- Terminate TLS at the Gateway for Web App connections when the GAT enables it: the connection goes through the same inner TLS upgrade Kubernetes connections use today, presenting the gateway certificate.
- Set `X-Forwarded-Proto` with the downstream scheme before forwarding the request
- Local dev tooling: the fake Twingate client gains a `WithDownstreamTLS()` option and `tools/local` runs an extra Web App client with it, testable via `curl --cacert ./test/data/proxy/tls.crt https://<client-address>`.


internal/token/gat_claims.go
```diff
@@ -82,7 +82,8 @@ func validatePort(port int, fieldName string) error {
 }
 
 func (p GATClaims) ShouldUpgradeTLS() bool {
-	return p.Resource.Type == ResourceTypeKubernetes
+	return p.Resource.Type == ResourceTypeKubernetes ||
+		(p.Resource.Type == ResourceTypeWebApp && p.Resource.GatewayMetadata.Downstream.TLS)
```
minhtule: nit: this is fine for now but as we support more protocols, a `switch` on the `p.Resource.Type` would be more readable.

internal/token/gat_claims.go
```diff
@@ -148,6 +149,8 @@ type GatewayMetadata struct {
 type Downstream struct {
 	// Port is the port that the protocol client connects to.
 	Port int `json:"port"`
+	// TLS indicates whether the Gateway terminates TLS for the protocol client.
```
minhtule: nit: it would be more like "TLS indicates whether the Gateway should **enforce** TLS for the protocol client." For HTTPS, the Gateway also does HTTPS redirect for non-HTTPS requests.

tools/local/main.go
```diff
@@ -160,6 +162,19 @@ func main() {
...
+		webAppGeo,
+		fmt.Sprintf("%s:%d", gatewayHost, gatewayPort),
+		controller.URL,
+		echoServer.address,
+		token.ResourceTypeWebApp,
+		fake.WithDownstreamTLS(),
+	)
+	defer webAppTLSClient.Close()
+
+	logger.Info("Web app (downstream TLS) fake Twingate client is serving at", zap.String("address", webAppTLSClient.Address))
```
minhtule: nit: "Web app HTTPS fake Twingate client...". Should change the other fake client to "Web app HTTP fake Twingate Client.

minhtule (approved):
LGTM 👍

sghiocel:
It seems that the gateway needs to add `X-Forwarded-Proto` to solve for this kind of issue:

Without `X-Forwarded-Proto`:
```
[1] Browser address bar:  https://grafana.acme.int/
[2] Gateway decrypts TLS, forwards upstream over plain HTTP

[3] App receives, verbatim:
      GET / HTTP/1.1
      Host: grafana.acme.int
      X-Forwarded-Proto: ""      <- the only way it can know
      r.TLS == nil ? true        <- its own leg is plaintext, by design

[4] App concludes scheme="http", replies:
      HTTP/1.1 302 Found
      Location: http://grafana.acme.int/login?redirect_uri=http://grafana.acme.int/callback

[5] Browser receives that Location
[6] Browser obeys it and goes to http://...
[7] That request to the Gateway returns: EOF
    ❌ LOGIN IS DEAD
```

With:
```
[3] App receives, verbatim:
      GET / HTTP/1.1
      Host: grafana.acme.int
      X-Forwarded-Proto: "https"    <- now it knows
      r.TLS == nil ? true           <- still plaintext, still by design

[4] App concludes scheme="https", replies:
      HTTP/1.1 302 Found
      Location: https://grafana.acme.int/login?redirect_uri=https://grafana.acme.int/callback

[6] Browser follows over HTTPS -> stays on the working path
    ✅ LOGIN WORKS
```

clement0010:
@sghiocel I added the `X-Forwarded-Proto` header in [66f15ac](https://github.com/Twingate/gateway/pull/422/commits/66f15acb51cad94a5d476da7b353f287278f6aee), could you review again?

sghiocel (approved):

Issue #423, opened 2026-07-24
Support multiple static certs and automated cert signing for downstream TLS


clement0010:
## Summary

The Gateway should serve a list of static TLS certificate and issue a new TLS cert when no static TLS certificate could be served for the SNI in the `ClientHello` message.

## Requirements
Update the Gateway config to support multiple static TLS certs and automated TLS cert signing 
```yaml
tls:
  certificates:
    files:
      - certificateFile: "/etc/gateway/tls/tls.crt"
        privateKeyFile: "/etc/gateway/tls/tls.key"
  automation:
    certificate:
      ttl: "24h"
      key:
        type: "ecdsa"
        bits: 256

    # There can be only 1 issuer specified.
    issuer:
      local:
         certificateFile: "/path/to/ca.crt"
         privateKeyFile: "/path/to/ca.key"

      vault:
         address: "<https://vault:8200>"
         namespace: ""
         caBundleFile: ""
         mount: "pki"
         role: "gateway"
         auth:
             token: "..."
             appRole: { mount: "approle", roleID: "...", secretID: "..." }
             gcp: { type: "gce|iam", role: "...", serviceAccountEmail: "..." }
             aws: { type: "iam|ec2", role: "..." }

      gcpPrivateCA:
        project: acme-prod 
        location: us-east1
        caPoolID: my-pool
        issuingCertificateAuthorityID: gateway-ca (Optional to pin issuing CA)
        # credentials are optional if workload identity is enabled
        credentialsFile: "/etc/gateway/gcp/credentials.json
```

The gateway look up for the static TLS cert that matches the SNI and serve it when found. Otherwise, it should generate and serve a new TLS cert with the configured issuer.

PR #424, opened 2026-07-24
feat(tls)!: Support multiple downstream static TLS certificates

clement0010:

- Issue: part of #423

## Changes

- Move the downstream TLS certificate settings from `tls.certificateFile`/`tls.privateKeyFile` to a `tls.certificates.files` block
  - Config validation now requires at least one TLS certificate to be present.
  - The certificate key pair file paths must be unique
- Update the Helm chart to render the new `tls.certificates.files` layout, so chart-managed deployments are unaffected.

## Notes

- **Breaking config schema change** for hand-written gateway configs: the two certificate fields must move under `tls.certificates.files`. Runtime behavior is unchanged, same certificate serving and hot reload.


internal/config/config.go
```diff
@@ -336,11 +341,25 @@ func (c *Config) Validate() error {
...
+	}
+
+	if err := t.Static.Validate(); err != nil {
+		return fmt.Errorf("static: %w", err)
+	}
+
+	return nil
+}
+
+var ErrMissingTLSConfig = errors.New("'static' must be specified for TLS config")
```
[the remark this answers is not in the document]
clement0010: Updated the `gateway-configmap.yaml`, so it should work as is

internal/config/config.go
```diff
@@ -87,7 +87,17 @@ type AuditLogConfig struct {
 	FlushSizeThreshold int           `yaml:"flushSizeThreshold"` // bytes
 }
 
+// TLSConfig represents the downstream TLS configuration.
 type TLSConfig struct {
+	Certificates TLSCertificatesConfig `yaml:"certificates"`
+}
```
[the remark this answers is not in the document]
clement0010: Updated it

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
+		logger:    logger,
+		files:     files,
+		certs:     make(map[string]*tls.Certificate, len(files)),
+		reloaders: make([]*reloader.Reloader, 0, len(files)),
+	}
+
+	for _, file := range files {
+		load := func() error { return cr.load(file) }
+		cr.reloaders = append(cr.reloaders, reloader.New([]string{file.CertificateFile, file.PrivateKeyFile}, load, logger))
 	}
```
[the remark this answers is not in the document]
clement0010: [Go1.22](https://go.dev/blog/loopvar-preview) should've fixed this

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
+	}
 }
 
-func (cr *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+// errNoCertificates is returned when no certificate has been loaded.
+var errNoCertificates = errors.New("no certificates configured")
+
+// GetCertificate returns the first certificate the client supports, falling back to the
+// first loaded certificate when none of them matches.
+func (cr *CertReloader) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
clement0010, line 55: This implementation is based on Go's [`getCertificate()`](https://github.com/golang/go/blob/e83070192eb33a282c75faa36457c4d1559f085f/src/crypto/tls/common.go#L1335-L1376) in `crypto/tls` module to preserve the behavior we see today

internal/config/config.go
```diff
@@ -89,7 +87,17 @@ type AuditLogConfig struct {
 	FlushSizeThreshold int           `yaml:"flushSizeThreshold"` // bytes
 }
 
+// TLSConfig represents the downstream TLS configuration.
 type TLSConfig struct {
+	Certificates TLSCertificatesConfig `yaml:"certificates"`
```
minhtule: A clearer name would be `TLSCertificateSources`.

internal/config/config.go
```diff
@@ -89,7 +87,17 @@ type AuditLogConfig struct {
...
+}
+
+// TLSCertificatesConfig lists the TLS certificates the Gateway can serve.
+type TLSCertificatesConfig struct {
+	Files []TLSCertificateFile `yaml:"files"`
+}
+
+type TLSCertificateFile struct {
```
minhtule: This struct has 2 files so I think `TLSCertificateFileKeyPair ` would be clearer. If we add `tls.certificates.pems` source later, that struct would be `TLSCertificatePEMKeyPair`.

internal/config/config.go
```diff
@@ -351,6 +358,20 @@ func (c *Config) Validate() error {
 }
 
 func (t *TLSConfig) Validate() error {
+	if len(t.Certificates.Files) == 0 {
+		return fmt.Errorf("%w: certificates.files", ErrRequired)
+	}
+
+	for i, file := range t.Certificates.Files {
```
minhtule: `pair` would be a clearer variable.

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
 
-	return cr.cert, nil
+	if len(cr.certs) == 0 {
+		return nil, errNoCertificates
+	}
+
+	var defaultCert *tls.Certificate
+
+	for _, file := range cr.files {
+		cert, ok := cr.certs[file.CertificateFile]
```
minhtule: I think it would be nicer to add a validation in the config to make sure that the certificate paths are unique.

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
+// GetCertificate returns the first certificate the client supports, falling back to the
+// first loaded certificate when none of them matches.
+func (cr *CertReloader) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
 	cr.mu.RLock()
 	defer cr.mu.RUnlock()
 
-	return cr.cert, nil
+	if len(cr.certs) == 0 {
+		return nil, errNoCertificates
+	}
```
minhtule: The config validation would prevent this from happening but it could still be possible that I configure 5 key pairs but none of them are valid. So it would be more robust to move this check to after the for loop. For example,
```go
for _, pair := range cr.files { 
    // as in today
}

if defaultCert == nil {
	return nil, errNoCertificates
}

// If nothing matches, return the first loaded certificate.
return defaultCert, nil
```
I would also change the error message to be
```go
var errNoCertificates = errors.New("no certificate could be loaded")
```

internal/connect/cert_reloader_test.go
```diff
@@ -99,28 +180,26 @@ func generateCert(t *testing.T) tls.Certificate {
 	}
 }
 
-func createCertFiles(t *testing.T, cert tls.Certificate) (certFile string, keyFile string) {
+func createCertFiles(t *testing.T, cert tls.Certificate) config.TLSCertificateFile {
```
minhtule: I would use `createKeyPairFiles` because Go's `tls` package uses "key pair" to refer to these 2 file ([doc](https://pkg.go.dev/crypto/tls#LoadX509KeyPair))

internal/config/config.go
clement0010: Do you mean `tls.certificates.directories`? I think the file format in `tls.certificates.files` today is already PEM format, and [`tls.LoadX509KeyPair()`](https://pkg.go.dev/crypto/tls#LoadX509KeyPair) expects it to be PEM

internal/connect/cert_reloader.go
clement0010: Good point, added in [939ba2e](https://github.com/Twingate/gateway/pull/424/commits/939ba2e81c508d965b9b00089972037a65fe2976). I later dropped private key validation in [e779853](https://github.com/Twingate/gateway/pull/424/commits/e779853e228114e03892da2727199657380b7bd6) since we could have different cert signed by the same private key

internal/connect/cert_reloader_test.go
clement0010: Renamed it to `createKeyPair` ([b2e1ee2](https://github.com/Twingate/gateway/pull/424/commits/b2e1ee2d23a3bb0a8d2893b74b15b4bb02e5b7df)) instead since [`tls.LoadX509KeyPair`](https://pkg.go.dev/crypto/tls#LoadX509KeyPair) doesn't have the `Files` suffix either

internal/connect/cert_reloader.go
clement0010: Updated in [75e3b12](https://github.com/Twingate/gateway/pull/424/commits/75e3b12e93d8335cc5d02ebe14ace0f8350ee072)

internal/connect/cert_reloader_test.go
```diff
@@ -17,79 +17,160 @@ import (
...
+		})
+	}
+}
+
+func clientHello(serverName string) *tls.ClientHelloInfo {
+	return &tls.ClientHelloInfo{
+		ServerName:        serverName,
+		SupportedVersions: []uint16{tls.VersionTLS13},
+	}
+}
```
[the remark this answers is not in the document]
clement0010: This is false, the check is skipped [here](https://github.com/golang/go/blob/e83070192eb33a282c75faa36457c4d1559f085f/src/crypto/tls/common.go#L1454-L1458) if the client doesn't send the signature schemes.

internal/config/config.go
minhtule: `tls.certificates.files` specifies the file paths in the `config.yaml`. `tls.certificates.pems` would put the cert and key's PEM content in the `config.yaml`.

internal/connect/listener_test.go
```diff
@@ -19,9 +19,17 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 
+	"gateway/internal/config"
 	"gateway/internal/token"
 )
 
+var testCertFiles = []config.TLSCertificateFileKeyPair{
```
minhtule: I would name this `testKeyPair` to be more aligned with the type

internal/config/config_test.go
```diff
@@ -357,8 +365,7 @@ func TestConfig_Validate(t *testing.T) {
 				Port:        8443,
 				MetricsPort: 9090,
 				TLS: TLSConfig{
-					CertificateFile: "tls.crt",
-					PrivateKeyFile:  "tls.key",
+					Certificates: TLSCertificateSources{Files: []TLSCertificateFileKeyPair{{CertificateFile: "tls.crt", PrivateKeyFile: "tls.key"}}},
```
minhtule: nit: I think it would be nicer to use multi line here

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
+	"gateway/internal/config"
 	"gateway/internal/reloader"
 )
 
 type CertReloader struct {
-	certFile string
-	keyFile  string
-	logger   *zap.Logger
+	logger *zap.Logger
+	files  []config.TLSCertificateFileKeyPair
```
minhtule: I would name this `keyPairs` instead. We're no longer in the context of the config so it's better to align the field name with the type.

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
-	cert *tls.Certificate
+	mu    sync.RWMutex
+	certs map[string]*tls.Certificate
 
-	reloader *reloader.Reloader
+	reloaders []*reloader.Reloader
 }
 
-func NewCertReloader(certFile, keyFile string, logger *zap.Logger) *CertReloader {
+func NewCertReloader(files []config.TLSCertificateFileKeyPair, logger *zap.Logger) *CertReloader {
```
minhtule: `keyPairs`

internal/connect/cert_reloader.go
```diff
@@ -6,57 +6,92 @@ package connect
...
 func (cr *CertReloader) Run(ctx context.Context) {
-	cr.reloader.Run(ctx)
+	for _, r := range cr.reloaders {
+		r.Run(ctx)
+	}
 }
 
-func (cr *CertReloader) GetCertificate(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+// errNoCertificates is returned when no certificate has been loaded.
+var errNoCertificates = errors.New("no certificate could be loaded")
```
minhtule: shouldn't error be at the top of the file?

internal/connect/cert_reloader_test.go
```diff
@@ -17,79 +17,160 @@ import (
...
+
+	"gateway/internal/config"
 )
 
 func TestCertReloader_Run(t *testing.T) {
-	cert := generateCert(t)
-	certFile, keyFile := createCertFiles(t, cert)
+	fooCert := generateCert(t, "foo.acme.int")
+	barCert := generateCert(t, "bar.acme.int")
+	fooFile := createKeyPair(t, fooCert)
```
minhtule: should be `fooKeyPair`?

internal/connect/cert_reloader_test.go
```diff
@@ -99,28 +180,26 @@ func generateCert(t *testing.T) tls.Certificate {
...
+		PrivateKeyFile:  filepath.Join(tmpDir, "tls.key"),
+	}
+	replaceCertFiles(t, file, cert)
 
-	return certFile, keyFile
+	return file
 }
 
-func replaceCertFiles(t *testing.T, certFile, keyFile string, newCert tls.Certificate) {
+func replaceCertFiles(t *testing.T, file config.TLSCertificateFileKeyPair, newCert tls.Certificate) {
```
minhtule: nit: I would name this `replaceKeyPairFiles` to mirror `createKeyPair`

internal/connect/cert_reloader_test.go
```diff
@@ -99,28 +180,26 @@ func generateCert(t *testing.T) tls.Certificate {
...
 	tmpDir := t.TempDir()
 
-	certFile = filepath.Join(tmpDir, "tls.crt")
-	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Certificate[0]})
-	require.NoError(t, os.WriteFile(certFile, certPEM, 0600))
-
-	keyFile = filepath.Join(tmpDir, "tls.key")
-	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(cert.PrivateKey.(*rsa.PrivateKey))})
-	require.NoError(t, os.WriteFile(keyFile, keyPEM, 0600))
+	file := config.TLSCertificateFileKeyPair{
```
minhtule: `keyPair` instead of `file`

internal/connect/cert_reloader_test.go
```diff
@@ -17,79 +17,160 @@ import (
...
-	requireCertReloader(t, cr, newCert)
+	requireCert(t, cr, "bar.acme.int", newBarCert)
+	requireCert(t, cr, "foo.acme.int", fooCert)
 }
 
 func TestCertReloader_load(t *testing.T) {
 	cert := generateCert(t)
-	certFile, keyFile := createCertFiles(t, cert)
-	_, otherKeyFile := createCertFiles(t, generateCert(t))
+	file := createKeyPair(t, cert)
```
minhtule: `keyPair`?

minhtule (approved):
Let's make sure all the variable names are consistent with the new type names. Otherwise, LGTM 👍

PR #425, opened 2026-07-24
feat(tls): support automated downstream certificate issuance

clement0010:

- Issue: #423

## Changes

- Support `tls.automation` config. When configured, Gateway can sign a TLS cert using the CA in `automation.issuer` based on the SNI from `ClientHello` in the outer TLS connection.
  - Cache the TLS cert by the SNI and sign a new one after the cert expires
  - Bound the maximum cache size at 1024 entries
  - Purge the cache when the issuer rotated its CA
- `tls.certificates` and `tls.automation` can be set together. A configured certificate covering the SNI is selected first. Only the uncovered host is issued on demand. At least one certificate needs to be supplied when no automation is configured.
- Support signing TLS cert using a local issuer

## Notes

- Accessing a web app resource by IP address over TLS does not work because `ClientHello` does not set IP address in SNI, recommend to use a DNS address.


internal/connect/cert.go
```diff
@@ -0,0 +1,225 @@
...
+	names = append(names, host)
+
+	for _, alias := range aliases {
+		if alias != "" && !slices.Contains(names, alias) {
+			names = append(names, alias)
+		}
+	}
+
+	return names
+}
```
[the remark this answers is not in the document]
clement0010: Fixed in a2e192f: certNames now sorts the aliases (host stays first as the CN), so the same name set always produces the same cache key regardless of token order.

internal/connect/cert/reloader.go
```diff
@@ -53,32 +53,50 @@ func (cr *CertReloader) Run(ctx context.Context) {
...
 		}
 
 		if err := hello.SupportsCertificate(cert); err == nil {
-			return cert, nil
-		}
-
-		if defaultCert == nil {
-			defaultCert = cert
+			return cert
 		}
```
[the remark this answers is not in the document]
clement0010: Not true, hostname is verified in [SupportCertificate(cert)](https://github.com/golang/go/blob/ac059878bcf58c4369fa7bef05bf34fb163b43e0/src/crypto/tls/common.go#L1406-L1414)

internal/connect/cert/manager.go
```diff
@@ -0,0 +1,75 @@
...
+// When the host is not covered by a configured certificate, it issues a new one on demand through the automation
+// and includes the aliases as additional subject alternative names.
+func (m *CertManager) GetCertificateForHost(hello *tls.ClientHelloInfo, host string, aliases ...string) (*tls.Certificate, error) {
+	if m.automation == nil {
+		return m.certs.GetCertificate(hello)
+	}
+
+	if cert := m.certs.MatchCertificate(hello); cert != nil {
+		return cert, nil
+	}
```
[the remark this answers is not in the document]
clement0010: This is not true, the SNI in `clienthello` has the same hostname as the requested host. So `MatchCertificate()` should return the correct certificate that covers the requested host.

internal/config/config.go
```diff
@@ -395,6 +427,72 @@ func (t *TLSCertificateFileKeyPair) Validate() error {
...
+		"ecdsa": true,
+		"rsa":   true,
+	}
+
+	if k.Type != "" && !validTypes[k.Type] {
+		return fmt.Errorf("%w: %q", ErrInvalidTLSKeyType, k.Type)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
clement0010: Validation is covered in [`keys.go`](https://github.com/Twingate/gateway/pull/425/changes#diff-58f2d575c92a39f66e6ad66be26e300b4e1e525f632a3c1885ae3e909eb3a418R33-R63)

internal/connect/cert_manager.go
```diff
@@ -0,0 +1,77 @@
...
+	if host == "" {
+		var err error
+
+		host, _, err = net.SplitHostPort(hello.Conn.LocalAddr().String())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse local address: %w", err)
+		}
+	}
+
+	return m.GetCertificateForHost(hello, host)
```
clement0010: For an IP address with only automation configured, we use `localAddr()` to issue the TLS cert. It will eventually fail the handshake.

internal/connect/cert_manager.go
```diff
@@ -0,0 +1,77 @@
...
+	if m.automation != nil {
+		m.automation.Run(ctx)
+	}
+}
+
+// GetCertificate returns a certificate for the outer TLS handshake based on the SNI host, falling back to
+// the connection's local IP for clients that send none (e.g. health probes).
+func (m *CertManager) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	host := hello.ServerName
+	if host == "" {
```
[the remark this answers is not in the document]
clement0010: This will be solved when we remove the outer TLS hostname verification

internal/connect/cert_manager.go
```diff
@@ -0,0 +1,77 @@
...
+		var err error
+
+		host, _, err = net.SplitHostPort(hello.Conn.LocalAddr().String())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse local address: %w", err)
+		}
+	}
+
+	return m.GetCertificateForHost(hello, host)
+}
```
clement0010: For an IP address with only automation configured, we use `localAddr()` to issue the TLS cert. So a resource with only an IP address configured would fail the outer TLS handshake.

internal/connect/cert_automation.go
```diff
@@ -0,0 +1,162 @@
...
+	// When the issuer can rotate its CA, drop the cached certificates as soon as
+	// it does rather than serving ones that no longer chain to it.
+	if issuer, ok := c.issuer.(rotatableIssuer); ok {
+		go c.purgeOnRotation(ctx, issuer.rotated())
+	}
+}
+
+func (c *CertAutomation) GetCertificateForHost(ctx context.Context, host string, aliases ...string) (*tls.Certificate, error) {
+	names := certNames(host, aliases)
+	key := strings.Join(names, ",")
```
clement0010: It is possible to have overlapping cache slots, e.g. 
Connect to address `bar.foo.int`
- Outer-layer TLS signs `bar.foo.int`
- Inner-layer TLS signs `*.foo.int` (GAT token address) or `bar.foo.int,alias.foo.int` (GAT token address + alias)

Outer TLS always sign a single hostname, I think this is unavoidable...

internal/connect/cert_manager.go
minhtule: Not sure I understand this. If we use `localAddr()` and the TLS is never usable, why don't we just fail? btw, what is the value of `hello.Conn.LocalAddr()`? I think it's the IP of the Gateway's listening socket.

internal/connect/cert_automation.go
minhtule: Can we use the outer TLS cert for `bar.foo.int` to serve the inner TLS?

internal/connect/cert/automation.go
minhtule, line 1: Instead of prefixing these files, I would keep them in `intern/connect/cert` directory. In the future, they could also be in their own package if we decide to only do this TLS automation for the inner layer.

internal/config/config.go
```diff
@@ -93,6 +93,7 @@ type AuditLogConfig struct {
 // TLSConfig represents the downstream TLS configuration.
 type TLSConfig struct {
 	Certificates TLSCertificateSources `yaml:"certificates"`
```
minhtule: If I specify `automation`, I shouldn't need to provide `certificates`. If the `Certificates` is optional now, should we convert this to pointer to be consistent with other optional fields?

internal/connect/cert_manager.go
clement0010: > If we use localAddr() and the TLS is still not usable, why don't we just fail?

We can't fail here because of the readiness and liveness probes, they connect without an SNI. I don't think kubelet checks the cert so the returned cert can be for any address. Perhaps we could hardcode a fixed address or pre-generate a cert for connections without SNI?

> what is the value of hello.Conn.LocalAddr() when the client is connecting using IP?

It should be the pod IP

internal/connect/cert_automation.go
clement0010: Yes, we can. This means we generate one TLS cert per SNI.

internal/connect/conn.go
```diff
@@ -246,6 +255,20 @@ func (p *ProxyConn) UpgradeToTLS() error {
...
 
+func (p *ProxyConn) getTLSConfig() *tls.Config {
+	tlsConfig := p.TLSConfig.Clone()
+	tlsConfig.GetCertificate = func(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+		cert, err := p.CertManager.GetCertificateForHost(hello, p.Claims.Resource.Address, p.Claims.Resource.Aliases...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get certificate for %q: %w", p.RequestedHost, err)
+		}
+
+		return cert, nil
```
minhtule: I think we could reuse the `p.TLSConfig` because Twingate Client should send the same SNI as the browser. That would avoid potentially 2 cert requests per connection.

internal/connect/cert_issuer.go
```diff
@@ -0,0 +1,194 @@
...
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+	}
+
+	for _, name := range names {
+		if ip := net.ParseIP(name); ip != nil {
+			template.IPAddresses = append(template.IPAddresses, ip)
+		} else {
+			template.DNSNames = append(template.DNSNames, name)
+		}
+	}
```
minhtule: This code seems to repeat in the Vault PR. I think a nicer abstraction would be
```
type certIssuer interface {
  sign(ctx context.Context, csr *x509.CertificateRequest) (cert *x509.Certificate, caChain []*x509.Certificate, error)
}
```
This issuer should work for any caller needing x509 certs. There should be coupled with `tls` package here.

internal/connect/cert_manager.go
minhtule: > Perhaps we could hardcode a fixed address or pre-generate a cert for connections without SNI?

I think the later is nicer. Or generate a cert without SAN here.

internal/connect/listener.go
```diff
@@ -102,12 +102,15 @@ func NewListener(
...
+	if err != nil {
+		return nil, err
+	}
 
 	tlsConfig := &tls.Config{
 		MinVersion:     tls.VersionTLS13,
 		MaxVersion:     tls.VersionTLS13,
-		GetCertificate: certReloader.GetCertificate,
+		GetCertificate: certManager.GetCertificate,
 	}
```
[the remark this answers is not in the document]
clement0010: known issue

internal/connect/cert/reloader.go
```diff
@@ -0,0 +1,115 @@
...
+	if cert := cr.firstCertificate(); cert != nil {
+		return cert, nil
+	}
+
+	return nil, ErrNoCertificates
+}
+
+// MatchCertificate returns the first certificate the client supports, or nil when
+// none of them matches.
+func (cr *reloader) MatchCertificate(hello *tls.ClientHelloInfo) *tls.Certificate {
```
minhtule: Does this need to be public?

internal/connect/conn_test.go
minhtule, line 1: Not sure we need these tests. `conn.go` doesn't change...

internal/config/config.go
```diff
@@ -395,6 +441,73 @@ func (t *TLSCertificateFileKeyPair) Validate() error {
 	return nil
 }
 
+var (
+	ErrMissingTLSCertificateSource = errors.New("at least one TLS certificate source must be configured")
```
minhtule: "automation" is not exactly a source. "certificates" has only `files` source type today. I would change this message to "either 'certificates' or 'automation' must be specified for TLS config".

This error is used above at line 392. We might want to move these error definitions up there.

internal/connect/cert/manager.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"crypto/tls"
+	"fmt"
+
+	"go.uber.org/zap"
+
+	"gateway/internal/config"
+)
+
+// Manager returns the certificate served on a downstream TLS handshake.
```
minhtule: In context of the gateway, "downstream" refers to the application's protocol downstream. TLS is also use in the CONNECT message between Client and Gateway.

internal/connect/cert/reloader.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (cr *reloader) Run(ctx context.Context) {
+	for _, r := range cr.reloaders {
+		r.Run(ctx)
+	}
+}
+
+// GetCertificate returns the first certificate the client supports, falling back to the
+// first loaded certificate when none of them matches.
+func (cr *reloader) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
minhtule: Looking only at this `reloader` struct, it's unclear why we need `GetCertificate` and `MatchCertificate`. This `GetCertificate` exists solely because of how the `Manager.GetCertificate` is written. I think this struct should only expose 2 methods
```go
type reloader struct {
  // return a certificate that could serves the client hello. Nil if there is none
  match(hello *tls.ClientHelloInfo) *tls.Certificate
  // return the first loaded certificate
  first() *tls.Certificate
}
```

Then the `Manager.GetCertificate()` logic would be.
1. A static cert the client supports wins.
2. Otherwise, with automation configured, issue for the SNI.
3. Otherwise, fall back to the first static cert.
4. Otherwise ErrNoCertificates.

For example,
```go
func (m *Manager) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
	if cert := m.certs.match(hello); cert != nil {
		return cert, nil
	}

	if m.automation != nil {
		cert, err := m.automation.GetCertificateForHost(hello.Context(), hello.ServerName)
		if err == nil {
			return cert, nil
		}

		m.logger.Warn("Failed to issue downstream certificate, falling back to a configured certificate",
			zap.String("host", hello.ServerName), zap.Error(err))
	}

	if cert := m.certs.first(); cert != nil {
		return cert, nil
	}

	return nil, ErrNoCertificates
}
```

Btw, this struct is private so all its methods should be lowercased.

internal/connect/cert/issuer.go
```diff
@@ -0,0 +1,235 @@
...
+	for _, ca := range caChain {
+		chain = append(chain, ca.Raw)
+	}
+
+	return &tls.Certificate{
+		Certificate: chain,
+		PrivateKey:  key,
+		Leaf:        leaf,
+	}, nil
+}
```
minhtule: Not sure if the refactor is still on-going but I think this logic would be reused by other issuers so it should be in the `automation`.

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return &automation{
+		issuer: issuer,
+		logger: logger,
+		cache:  cache,
+	}, nil
+}
+
+func (c *automation) Run(ctx context.Context) error {
```
minhtule: The struct is private so its methods should be private too?

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,149 @@
...
+	cert, ok := c.cache.Get(key)
+	if !ok {
+		return nil, false
+	}
+
+	lifetime := cert.Leaf.NotAfter.Sub(cert.Leaf.NotBefore)
+	renewAfter := cert.Leaf.NotBefore.Add(time.Duration(float64(lifetime) * renewFraction))
+
+	return cert, time.Now().Before(renewAfter)
+}
```
minhtule: There are 2 issues with this method:
- If we don't renew the cert, why don't we just use the cert's expiration?
- If the cert is not usable (passing its renewal timestamp currently), it would be better to return nil

I would change this method to just return a cert if there is a usable cert and nil otherwise.

internal/connect/cert/issuer.go
```diff
@@ -0,0 +1,235 @@
...
+
+	der, err := x509.CreateCertificateRequest(rand.Reader, template, key)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create certificate request: %w", err)
+	}
+
+	csr, err := x509.ParseCertificateRequest(der)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse certificate request: %w", err)
+	}
```
minhtule: It's a bit off that we need to parse the certificate request again. Perhaps, the `issuer` interface should have the `csr` argument as `[]byte`? 🤔 If we really need to, we could create a struct like cert-manager [`CertificateRequestObject`](https://github.com/cert-manager/issuer-lib/blob/main/controllers/signer/certificate_request_object.go#L48-L55).

clement0010: I think we should pass in a struct `certificateRequest` instead because I realized we need the details to verify the certificate returned by third-party CAs.

We should also have a `certificateRequest.csr()` method to return a CSR in `[]byte` because the Vault's sign API requires a PEM-encoded CSR ([docs](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-certificate)) and GCP Private CA's `pemCsr` field ([docs](https://docs.cloud.google.com/certificate-authority-service/docs/reference/rest/v1/projects.locations.caPools.certificates#Certificate)).

internal/connect/cert/automation.go
clement0010: Simplified in [`cf2a375`](https://github.com/Twingate/gateway/pull/425/commits/cf2a3752121f04c2905ce41162afc91b6a5c4b72)

internal/connect/cert/automation.go
clement0010: Fixed in [`f922b0d`](https://github.com/Twingate/gateway/pull/425/commits/f922b0d7f507097a5a190e1854e0335ee9bca0fc)

internal/connect/cert/reloader.go
clement0010: Fixed in [f922b0d](https://github.com/Twingate/gateway/pull/425/commits/f922b0d7f507097a5a190e1854e0335ee9bca0fc)

internal/config/config.go
clement0010: Fixed in [41747e2](https://github.com/Twingate/gateway/pull/425/commits/41747e231a7db9d5594d6f40b0fb650544e61783)

internal/connect/cert/issuer.go
clement0010: Yep, I am aware that this pattern can be refactored and placed in `automation.go`. Updated this in [`94e3ef0`](https://github.com/Twingate/gateway/pull/425/commits/94e3ef0e11715adee507892776422791b158328a) to drop `issue()` from `Issuer` interface.

internal/connect/conn_test.go
clement0010: Removed in [`38f2d3c`](https://github.com/Twingate/gateway/pull/425/commits/38f2d3c8d324eae3c71c2e807e9d1382f15b79f3)

internal/config/config.go
clement0010: Agree, lets convert it to a pointer to be consistent with other optional fields. This was updated in [2112032](https://github.com/Twingate/gateway/pull/425/commits/2112032e380f47373e02ee9fd375ffc4e98191a7)

internal/connect/conn.go
clement0010: Agree, lets convert it to a pointer to be consistent with other optional fields. This was updated in [2112032](https://github.com/Twingate/gateway/pull/425/commits/2112032e380f47373e02ee9fd375ffc4e98191a7)

internal/connect/cert/reloader.go
clement0010: Thank you for the suggestions, I adopted it in [`8e47edf`](https://github.com/Twingate/gateway/pull/425/commits/8e47edf91787b51dabb96c612ced7fe4e2adfebe)

internal/connect/cert/manager.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func NewManager(tlsCfg config.TLSConfig, logger *zap.Logger) (*Manager, error) {
+	var keyPairs []config.TLSCertificateFileKeyPair
+	if tlsCfg.Certificates != nil {
+		keyPairs = tlsCfg.Certificates.Files
+	}
+
+	manager := &Manager{certs: newReloader(keyPairs, logger)}
+
```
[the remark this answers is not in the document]
clement0010: Not related to this PR: if we want to fix it, best to fix it separately

internal/connect/cert/reloader.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (r *reloader) run(ctx context.Context) {
+	for _, fr := range r.reloaders {
+		fr.Run(ctx)
+	}
+}
+
+// match returns the first certificate in configuration order that the client supports.
+func (r *reloader) match(hello *tls.ClientHelloInfo) *tls.Certificate {
```
minhtule: The "the client supports" sounds off. It should be something like "the first certificate in configuration order that can serve this ClientHello."

internal/connect/cert/manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+			tlsCfg:     config.TLSConfig{Automation: testAutomationConfig()},
+			hello:      clientHello("app.acme.int"),
+			wantIssued: []string{"app.acme.int"},
+		},
+		{
+			name:       "IP SNI is issued on demand",
+			tlsCfg:     config.TLSConfig{Certificates: files, Automation: testAutomationConfig()},
+			hello:      clientHello("10.0.0.5"),
+			wantIssued: []string{"10.0.0.5"},
+		},
```
minhtule: Is it possible to have the IP in the TLS's client hello SNI?

internal/connect/cert/manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+			tlsCfg:   config.TLSConfig{Certificates: files, Automation: testAutomationConfig()},
+			hello:    clientHello(""),
+			wantCert: fooCert.Certificate,
+		},
+		{
+			name:       "SNI is issued on demand when no certificates are configured at all",
+			tlsCfg:     config.TLSConfig{Automation: testAutomationConfig()},
+			hello:      clientHello("app.acme.int"),
+			wantIssued: []string{"app.acme.int"},
+		},
```
minhtule: This case seems to cover the 2nd case "uncovered SNI is issued on demand"? 🤔

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	chain := make([][]byte, 0, len(caChain)+1)
+	chain = append(chain, leaf.Raw)
+
+	for _, ca := range caChain {
+		chain = append(chain, ca.Raw)
+	}
```
minhtule: nit: this is shorter and as readable.
```go
chain := make([][]byte, len(caChain)+1)
chain[0] = leaf.Raw

for i, ca := range caChain {
    chain[i+1] = ca.Raw
}
```

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,210 @@
...
+	key         crypto.Signer
+	dnsNames    []string
+	ipAddresses []net.IP
+	ttl         time.Duration
+}
+
+func newCertificateRequest(key crypto.Signer, name string, ttl time.Duration) *certificateRequest {
+	req := &certificateRequest{key: key, ttl: ttl}
+
+	// A handshake without SNI leaves no name, and the certificate then covers none.
```
minhtule: There is no SNI in this context. Should be something like "If `name` is empty, the cert has no SAN." It should probably a method's doc string.

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,210 @@
...
+	if ip := net.ParseIP(name); ip != nil {
+		req.ipAddresses = []net.IP{ip}
+	} else if name != "" {
+		req.dnsNames = []string{name}
+	}
+
+	return req
+}
+
+// csr builds the request DER-encoded and signed by the leaf key.
```
minhtule: Not sure about "leaf key" here. This make a CSR using a private key and get back a leaf cert. I would probably just drop the 2nd part of the comment.

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,210 @@
...
+	// it does rather than serving ones that no longer chain to it.
+	if issuer, ok := a.issuer.(rotatableIssuer); ok {
+		go a.purgeOnRotation(ctx, issuer.rotated())
+	}
+
+	return nil
+}
+
+// getCertificateForHost issues a certificate covering the given host, caching it under that name.
+func (a *automation) getCertificateForHost(ctx context.Context, host string) (*tls.Certificate, error) {
```
minhtule: nit: I would drop `ForHost` here. Other `getCertificate` methods don't need prepositions. The argument name and type should provide that context.

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+func (a *automation) cachedCert(key string) *tls.Certificate {
+	cert, ok := a.cache.Get(key)
+	if !ok {
+		return nil
+	}
+
+	if time.Now().Before(cert.Leaf.NotAfter) {
```
minhtule: Should add some buffer so we don't return a cert that is about to expire in 1s.

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,517 @@
...
+			local:   &config.TLSLocalIssuerConfig{CertificateFile: "../../../test/data/proxy/tls.crt", PrivateKeyFile: "../../../test/data/proxy/tls.key"},
+			key:     config.TLSCertificateKeyConfig{Type: "ed25519"},
+			wantErr: errUnsupportedKeyType,
+		},
+		{
+			name:    "unsupported key bits",
+			local:   &config.TLSLocalIssuerConfig{CertificateFile: "../../../test/data/proxy/tls.crt", PrivateKeyFile: "../../../test/data/proxy/tls.key"},
+			key:     config.TLSCertificateKeyConfig{Type: "ecdsa", Bits: 128},
+			wantErr: errUnsupportedKeyBits,
+		},
```
minhtule: These 2 tests repeat `newKeyConfig`'s tests. We should only need 1 method checking the case when `newKeyConfig` returns an error.

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,517 @@
...
+	_, err = issued.Leaf.Verify(x509.VerifyOptions{
+		Roots:     pool,
+		KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+	})
+	require.NoError(t, err, "leaf should still chain to the CA")
+
+	cached, ok := cert.cache.Get("")
+	require.True(t, ok, "the nameless certificate should be cached under the empty host")
+	assert.Same(t, issued, cached)
+}
```
minhtule: These 2 tests could collapse into 1 with table.

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,517 @@
...
+	cert := newStubAutomation(t, newStubIssuer())
+	cert.key = keyConfig{typ: keyTypeECDSA, bits: 128}
+
+	_, err := cert.issue(t.Context(), "app.acme.int")
+	require.ErrorIs(t, err, errUnsupportedKeyBits)
+	assert.Contains(t, err.Error(), "failed to generate leaf key")
+}
+
+func TestAutomation_getCertificateForHost(t *testing.T) {
+	cert, err := newAutomation(testAutomationConfig(), zap.NewNop())
```
minhtule: I think we could simplify the tests with using `stubAutomation` except the 2 tests for e2e with the local issuer
- `TestAutomation_issue` - verify cert is issued correct
- `TestAutomation_Run_ReissuesAfterCARotation` - verify rotation

Other tests should focus on the `automation`'s logic e.g. caching, handle expiration, etc. They shouldn't need to verify the issued certs.

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,517 @@
...
+			// A backend reads the names off the request but forwards the CSR, so the two
+			// have to agree, and the CSR has to prove possession of the key it asks for.
+			assert.Equal(t, requestNames(req), csrNames(csr))
+			assert.Empty(t, csr.Subject.CommonName)
+			assert.NoError(t, csr.CheckSignature())
+		})
+	}
+}
+
+func csrNames(csr *x509.CertificateRequest) []string {
```
minhtule: "name" is a bit vague. `csrSANs` would be better in X509 cert context.

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,517 @@
...
+	names := slices.Clone(csr.DNSNames)
+	for _, ip := range csr.IPAddresses {
+		names = append(names, ip.String())
+	}
+
+	return names
+}
+
+func requestNames(req *certificateRequest) []string {
```
minhtule: Ditto

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,517 @@
...
+	errStubRun  = errors.New("stub issuer failed to start")
+	errStubSign = errors.New("stub issuer failed to sign")
+)
+
+// stubIssuer signs placeholder certificates, reporting each signing start on entered
+// and blocking requests for a host that has a gate until their gate closes. Configure
+// gates before the first request.
+type stubIssuer struct {
+	entered  chan string
+	gates    map[string]chan struct{}
```
minhtule: These value are so generic and hard to understand even after reading the implementation. I would add a `onSign()` hook and for each test that need to hold the singing, they can set up the channels locally.

internal/connect/cert/manager_test.go
clement0010: I think it's possible but it's a non-standard practice and won't happen in our case. I removed it regardless

internal/connect/cert/manager_test.go
clement0010: The only difference is that there are no cert files configured here, the code path tested should be similar. I'll remove it

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,212 @@
...
+	cert, ok := a.cache.Get(key)
+	if !ok {
+		return nil
+	}
+
+	if time.Until(cert.Leaf.NotAfter) > expiryBuffer {
+		return cert
+	}
+
+	return nil
```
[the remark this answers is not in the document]
clement0010: We don't have to remove it since the new cert would overwrite it. It might also introduce a problem where another goroutine removes a newly issued cert.

internal/connect/cert/automation_test.go
```diff
@@ -0,0 +1,496 @@
...
+
+	close(start)
+	wg.Wait()
+
+	assert.Equal(t, 1, cert.cache.Len())
+}
+
+func TestAutomation_getCertificate_SlowIssuanceDoesNotBlockOtherHosts(t *testing.T) {
+	started := make(chan struct{}, 1)
+	block := make(chan struct{})
```
minhtule: `started` and `block` describe the mechanism (a channel that gets a send, a channel that blocks) rather than the event each one marks, so the reader has to trace the goroutines to learn what they mean. Better names would be `slowSigning` and `slowRelease`.

minhtule (approved):
LGTM 👍

internal/connect/cert/issuer.go
```diff
@@ -0,0 +1,174 @@
...
+
+	caCert, err := x509.ParseCertificate(pair.Certificate[0])
+	if err != nil {
+		return fmt.Errorf("failed to parse CA certificate: %w", err)
+	}
+
+	if !caCert.IsCA {
+		return fmt.Errorf("%w: %q", errNotCACertificate, l.certFile)
+	}
```
minhtule: I think this error could be better.
```go
return fmt.Errorf("%w: %q must have basicConstraints CA:TRUE", errNotCA, l.certFile)
```
This cert's key usage should also have cert signing.

internal/connect/cert/issuer.go
```diff
@@ -0,0 +1,174 @@
...
+	"go.uber.org/zap"
+
+	"gateway/internal/config"
+	filereloader "gateway/internal/reloader"
+)
+
+const clockSkewBuffer = 30 * time.Second
+
+var (
+	errNotCACertificate = errors.New("certificate is not a certificate authority")
```
minhtule: nit: "certificate is not a CA certificate"

internal/connect/cert/issuer.go
clement0010: Added a stricter validation in [`0404cc1`](https://github.com/Twingate/gateway/pull/425/commits/0404cc14273356febc5d92dc2b07973e47e9af2b), I took the implementation from Go's [x509](https://github.com/golang/go/blob/e51216de8e26247ee0f3d2cfa576233b0d29f542/src/crypto/x509/x509.go#L1020-L1032)

internal/connect/cert/automation.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+func (a *automation) cachedCert(key string) *tls.Certificate {
+	cert, ok := a.cache.Get(key)
+	if !ok {
+		return nil
+	}
+
+	if time.Until(cert.Leaf.NotAfter) > expiryBuffer {
```
sghiocel, line 174: If the config's TTL is <= the 5 minute `expiryBuffer`, we lose the ability to cache here. Should we enforce a 10 minute minimum TTL in the config so that we can't run into this situation?

clement0010: Yes, I think we should. Added in [`bf13ef2`](https://github.com/Twingate/gateway/pull/425/commits/bf13ef243bb9282bc016ab9c093f28cfe215a837)

sghiocel (approved):

PR #437, opened 2026-07-29
feat(helm): support specifying the Twingate Remote Network

minhtule:

- Issue: https://github.com/Twingate/gateway/issues/330
- Requires Twingate/kubernetes-operator#1109

## Changes

- `twingateOperator.remoteNetworkId`: Remote Network the `TwingateGateway` and the
  in-cluster Kubernetes `TwingateResource` are created in
  - empty (default) falls back to the operator's `TWINGATE_REMOTE_NETWORK_ID`
- error when `twingateOperator.kubernetesResource.remoteNetworkId` is set

## Notes

- One shared key rather than one per CR: Twingate requires the resource to be in
  its gateway's Remote Network, and the gateway can be rendered without the
  resource. Setting it under `kubernetesResource` fails rather than being
  silently overridden by the shared value.
- The operator CRD makes the field immutable once set, so changing the value on
  an existing release is rejected on upgrade.

### DIFF
--- a/deploy/gateway/tests/twingate-cr_test.yaml
+++ b/deploy/gateway/tests/twingate-cr_test.yaml
@@ -14,6 +14,17 @@ tests:
     asserts:
       - failedTemplate:
           errorMessage: twingateOperator.kubernetesResource.enabled=true requires twingateOperator.gateway.enabled=true
+  - it: fails when the in-cluster resource sets its own remote network
+    set:
+      twingateOperator:
+        gateway:
+          enabled: true
+        kubernetesResource:
+          enabled: true
+          remoteNetworkId: "UmVtb3RlTmV0d29yazo0NTYK"
+    asserts:
+      - failedTemplate:
+          errorMessage: twingateOperator.kubernetesResource.remoteNetworkId is not supported, set twingateOperator.remoteNetworkId instead
   - it: renders the CA and Gateway when gateway is enabled
     set:
       twingateOperator:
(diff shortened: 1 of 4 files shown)


deploy/gateway/values.yaml
```diff
@@ -10,6 +10,9 @@ twingate:
   network: ""
 
 twingateOperator:
+  # Remote Network the Gateway and the in-cluster Kubernetes resource are created in.
+  # Empty uses the operator's own TWINGATE_REMOTE_NETWORK_ID.
+  remoteNetworkId: ""
```
ekampf: Why here and not under gateway?

minhtule: because it would apply on both the `gateway` and `kubernetesResource`

deploy/gateway/tests/twingate-cr_test.yaml
```diff
@@ -62,6 +73,47 @@ tests:
...
+      twingateOperator:
+        gateway:
+          enabled: true
+        kubernetesResource:
+          enabled: true
+    kubernetesProvider: *clusterNamespace
+    asserts:
+      - notExists:
+          path: spec.remoteNetworkId
+  - it: sets remoteNetworkId on the gateway
```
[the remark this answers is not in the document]
minhtule: The test does select the `TwingateGateway`

deploy/gateway/values.yaml
ekampf: lets call it `remoteNetworkIdOverride`

ekampf (approved):

deploy/gateway/values.yaml
minhtule: Discussed offline and decided to keep `remoteNetworkId` name but comment out this field.

PR #440, opened 2026-08-03
feat: support Vault as TLS automation cert issuer

clement0010:
## Related Tickets

Issue: #423

## Changes

- Support `tls.automation.issuer.vault` config. When configured, Gateway can issue a TLS cert via Vault through the `/sign` endpoint.
  - The issued TLS cert is verified to contain the exact requested alternative names and IP sans, TTL and the public key.
- Rename the shared Vault auth config structs (`SSHCAVault*` → `Vault*`) for reuse by both SSH and TLS configs; the YAML config schema is unchanged
- Support the same Vault connection and auth options as the SSH CA integration: token, AppRole, GCP, and AWS auth, with `VAULT_TOKEN` fallback and automatic token renewal
- Add an integration test to test the Vault PKI signing endpoint


internal/connect/cert.go
```diff
@@ -238,6 +242,122 @@ func (s *selfSignIssuer) issue(_ context.Context, names []string) (*tls.Certific
...
+
+	if len(altNames) > 0 {
+		data["alt_names"] = strings.Join(altNames, ",")
+	}
+
+	if len(ipSANs) > 0 {
+		data["ip_sans"] = strings.Join(ipSANs, ",")
+	}
+
+	secret, err := v.vault.Client.Logical().WriteWithContext(ctx, v.mount+"/issue/"+v.role, data)
```
minhtule: I think the Gateway should create the key and use the `/sign` endpoint to get the certificate so the key never leaves the Gateway.

internal/connect/cert.go
```diff
@@ -238,6 +242,122 @@ func (s *selfSignIssuer) issue(_ context.Context, names []string) (*tls.Certific
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to issue certificate: %w", err)
+	}
+
+	if secret == nil || secret.Data == nil {
+		return nil, fmt.Errorf("%w: empty response", errVaultIssueFailed)
+	}
+
+	certPEM, ok := secret.Data["certificate"].(string)
```
minhtule: Like SSH, we probably need to validate the the cert returned by the CA e.g. SANs are what we request.

internal/config/config.go
```diff
@@ -182,11 +190,7 @@ type SSHCAManualConfig struct {
 
 // SSHCAVaultConfig configures CAs backed by Vault's SSH secrets engine.
 type SSHCAVaultConfig struct {
-	Address      string               `yaml:"address"`                // Vault server address, e.g. https://vault.example.com:8200
-	CABundleFile string               `yaml:"caBundleFile,omitempty"` // Path to a PEM CA bundle for verifying Vault's TLS certificate; omit to use the system trust store
-	Auth         SSHCAVaultAuthConfig `yaml:"auth"`
-
-	Namespace string `yaml:"namespace,omitempty"` // Optional Vault namespace
```
clement0010, line 189: These are moved to `VaultConfig`

internal/config/config_test.go
```diff
@@ -304,7 +304,7 @@ ssh:
       ttl: "5m"
   ca:
     vault:
-      server: "https://vault:8200"
```
clement0010, line 308: Should be `address` typo from previous commit

internal/connect/listener.go
```diff
@@ -139,15 +139,17 @@ func NewListener(
...
 
 	defer func() {
 		wg.Wait()
 		l.closeChannels()
 	}()
 
+	if err := l.certManager.Run(ctx); err != nil {
+		return err
+	}
+
```
clement0010, line 153: Moving this here so we can close the channel when error

internal/connect/cert_issuer.go
```diff
@@ -192,3 +203,214 @@ func (l *localIssuer) issue(_ context.Context, names []string) (*tls.Certificate
...
+	}
+
+	csr, err := certificateRequestPEM(key, names)
+	if err != nil {
+		return nil, err
+	}
+
+	data := map[string]any{
+		"csr":         csr,
+		"common_name": names[0],
```
clement0010: I think Vault PKI role might reject the IP address as CN, so additional configuration might be needed to allow signing IP address as CN

minhtule: Can we simplify by not setting the common name? Browsers don't check for common name and cert-manager discourages setting it ([doc](https://cert-manager.io/docs/usage/certificate/#creating-certificate-resources)). Vault PKI role have the [`require_cn`](https://developer.hashicorp.com/vault/api-docs/secret/pki#require_cn) option to not require common name.

internal/sshhandler/ca.go
```diff
@@ -18,9 +18,10 @@ import (
 	"go.uber.org/zap"
 	"golang.org/x/crypto/ssh"
 
-	vault "github.com/hashicorp/vault/api"
```
minhtule, line 21: nit: The test file `ca_test.go` still uses this old import. Would be more consistent to change it to `vaultapi`.

internal/vault/vault.go
```diff
@@ -161,40 +162,59 @@ func newVault(vaultConfig *gatewayconfig.SSHCAVaultConfig, logger *zap.Logger) (
...
+		secret = v.LoginWithRetry(ctx)
 		if ctx.Err() != nil {
 			return
 		}
 	}
 }
 
-func (v *Vault) watchTokenLifecycle(ctx context.Context, secret *vault.Secret) error {
-	watcher, err := v.client.NewLifetimeWatcher(&vault.LifetimeWatcherInput{
+func (v *Vault) LoginWithRetry(ctx context.Context) *vaultapi.Secret {
```
minhtule: Why does this need to be public? Also why was it moved?

test/integration/tls_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+				return gatewayconfig.VaultAuthConfig{
+					AppRole: &gatewayconfig.VaultAppRoleConfig{
+						RoleID:   roleID,
+						SecretID: secretID,
+					},
+				}
+			},
+		},
+	}
```
minhtule, line 66: We can do it later but I think it's better to have a dedicate test for Vault auth. When we support a new auth method, we shouldn't have to add them to both this and SSH test.

test/integration/testutil/vault.go
```diff
@@ -105,6 +132,9 @@ path "ssh/sign/*" {
 path "ssh/config/ca" {
   capabilities = ["read"]
 }
+path "pki/sign/*" {
+  capabilities = ["create", "update"]
+}
```
minhtule, line 103: I would slit this into `SetupVaultSSHCA` and `SetupVaultPKI` to set up what is needed for each test. They both could reuse the common setup in this helper.

internal/connect/cert/issuer.go
```diff
@@ -172,3 +183,164 @@ func (l *localIssuer) sign(_ context.Context, req *certificateRequest) (*x509.Ce
...
+
+	for _, ip := range ips {
+		names[ip.String()] = struct{}{}
+	}
+
+	return names
+}
+
+// parseCertificateChain parses the PEM certificates in a CA's response, leaf first.
+func parseCertificateChain(pems []string) ([]*x509.Certificate, error) {
```
minhtule: This method parse a slice of certs. It doesn't know if the leaf cert is first or last. It also doesn't care if the list is from the CA's response or not.

internal/connect/cert/issuer.go
```diff
@@ -172,3 +183,164 @@ func (l *localIssuer) sign(_ context.Context, req *certificateRequest) (*x509.Ce
...
+
+		chain = append(chain, cert)
+	}
+
+	return chain, nil
+}
+
+// caChain returns the CA chain PEMs from the response, falling back to the
+// issuing CA when the chain is absent.
+func caChain(data map[string]any) []string {
```
minhtule: This function is specific to Vault so we should make it explicit e.g. `vaultCAChainPEMs` and explain what `data` argument is e.g. `"returns the CA chain PEMs from a Vault PKI sign response's Data, preferring ca_chain and falling back to issuing_ca"`

internal/connect/cert/issuer.go
```diff
@@ -172,3 +183,164 @@ func (l *localIssuer) sign(_ context.Context, req *certificateRequest) (*x509.Ce
...
+	if leaf.NotAfter.After(maxNotAfter) {
+		return fmt.Errorf("%w: validity %s exceeds requested TTL (max %s)", errIssuedCertMismatch, leaf.NotAfter, maxNotAfter)
+	}
+
+	return nil
+}
+
+// certificateNames returns the names as a set, lowercased and in canonical IP
+// notation so both sides of a comparison match.
+func certificateNames(dnsNames []string, ips []net.IP) map[string]struct{} {
```
minhtule: nit: I think `certificateSANs` are more accurate.

internal/connect/cert/issuer_test.go
```diff
@@ -188,11 +206,373 @@ func TestLocalIssuer_sign_CAKeyFails(t *testing.T) {
...
+
+func csrPEM(csr []byte) string {
+	return string(pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csr}))
+}
+
+func caPEM(t *testing.T, ca tls.Certificate) string {
+	t.Helper()
+
+	return string(pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: ca.Certificate[0]}))
+}
```
minhtule, line 301: why does `caPEM` have `t.Helper()` but `csrPEM` doesn't?

internal/connect/cert/issuer_test.go
```diff
@@ -188,11 +206,373 @@ func TestLocalIssuer_sign_CAKeyFails(t *testing.T) {
...
-
-	req := newCertificateRequest(key, "app.acme.int", defaultTTL)
+	req := newCertificateRequest(generateKey(t), "app.acme.int", defaultTTL)
 
 	_, _, err = issuer.sign(t.Context(), req)
 	require.ErrorContains(t, err, "failed to sign leaf certificate")
 }
+
+// signTestCSR signs the PEM CSR against a third-party CA signing endpoint.
+func signTestCSR(t *testing.T, ca tls.Certificate, csrPEM string) string {
```
minhtule: Where is the signing endpoint in this function?

internal/connect/cert/issuer_test.go
```diff
@@ -188,11 +206,373 @@ func TestLocalIssuer_sign_CAKeyFails(t *testing.T) {
...
+}
+
+func caPEM(t *testing.T, ca tls.Certificate) string {
+	t.Helper()
+
+	return string(pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: ca.Certificate[0]}))
+}
+
+// decodeSignPayload returns the request payload with the dynamic CSR split off.
+func decodeSignPayload(t *testing.T, r *http.Request) (csr string, payload map[string]any) {
```
minhtule: `decodeVaultSignPayload`

internal/connect/cert/issuer_test.go
```diff
@@ -188,11 +206,373 @@ func TestLocalIssuer_sign_CAKeyFails(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		pems    []string
+		wantLen int
+		wantErr error
+	}{
+		{
+			name:    "leaf first, then the CA",
+			pems:    []string{signTestCSR(t, ca, csrPEM(der)), caPEM(t, ca)},
+			wantLen: 2,
```
minhtule: If I put the CA cert before the leaf, does this test fail?

clement0010: This test is not about ordering, renamed to `certificate chain with multiple PEM blocks`

internal/connect/cert/issuer_test.go
clement0010: It's being called in a test vault server's signing endpoint but it could also sign without a signing endpoint. Removed the comment

internal/connect/cert/issuer_test.go
clement0010: Added `t.Helper()` to `csrPEM()`

internal/vault/vault.go
clement0010: Reverted back

test/integration/testutil/vault.go
clement0010: Updated in [`f6e4f31`](https://github.com/Twingate/gateway/pull/440/commits/f6e4f31dbc40368cd32c199f5395160faa279284)

internal/connect/cert/issuer.go
```diff
@@ -172,3 +183,152 @@ func (l *localIssuer) sign(_ context.Context, req *certificateRequest) (*x509.Ce
...
+	csr, err := req.csr()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	data := map[string]any{
+		"csr": string(pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csr})),
+		"ttl": req.ttl.String(),
+		// pem_bundle returns the CA chain concatenated onto the leaf
+		"format": "pem_bundle",
```
clement0010, line 244: I think it'd be simpler to use [`pem_bundle`](https://developer.hashicorp.com/vault/api-docs/secret/pki#pem_bundle-2) to let Vault return the certificate chains. So we avoid constructing the chain ourselves with `vaultCAChainPEMs()`

internal/connect/cert/issuer_test.go
minhtule: Why do we need `t.Helper()`?

test/integration/testutil/vault.go
```diff
@@ -17,7 +17,93 @@ import (
...
+path "ssh/config/ca" {
+  capabilities = ["read"]
+}`)
+
+	return containerName, serverPort
+}
+
+// SetupVaultPKI starts a Vault server holding the PKI CA that issues downstream TLS
+// certificates.
```
minhtule: Vault PKI issues X509 cert in general. Whether that cert is used for TLS or downstream is the caller's responsibility. We should avoid adding comments explaining how a function/struct is used.

internal/connect/cert/issuer.go
```diff
@@ -172,3 +183,152 @@ func (l *localIssuer) sign(_ context.Context, req *certificateRequest) (*x509.Ce
...
+		return fmt.Errorf("failed to login to Vault: %w", err)
+	}
+
+	go v.vault.RunTokenRenewalLoop(ctx, secret)
+
+	return nil
+}
+
+// sign has Vault sign the request through <mount>/sign/<role>, so the private key
+// never leaves the Gateway. The context is the TLS handshake's, so an abandoned
```
minhtule: I don't think we need to mention `so the private key never leaves the Gateway" in this interface.

internal/connect/cert/issuer.go
```diff
@@ -172,3 +183,152 @@ func (l *localIssuer) sign(_ context.Context, req *certificateRequest) (*x509.Ce
...
+	if err := verifyIssuedCertificate(leaf, req); err != nil {
+		return nil, nil, err
+	}
+
+	return leaf, chain[1:], nil
+}
+
+// verifyIssuedCertificate rejects a CA-issued certificate that is signed by a different signer or
+// grants more than the request asked for, in names or in validity.
+func verifyIssuedCertificate(leaf *x509.Certificate, req *certificateRequest) error {
```
minhtule: There is no signer here... A better comment would be
```
// verifyIssuedCertificate rejects a certificate whose public key, SANs, or validity differ from the request.
```

internal/connect/cert/issuer_test.go
```diff
@@ -188,11 +206,384 @@ func TestLocalIssuer_sign_CAKeyFails(t *testing.T) {
...
+			name: "extra granted name",
+			mutate: func(leaf *x509.Certificate) {
+				leaf.DNSNames = append(leaf.DNSNames, "extra.acme.int")
+			},
+			wantErr: true,
+		},
+		{
+			name: "extra granted IP",
+			mutate: func(leaf *x509.Certificate) {
+				leaf.IPAddresses = append(leaf.IPAddresses, net.ParseIP("10.0.0.6"))
```
minhtule: I remove this line and the test still passes. In general, we shouldn't mutation a setup across table test cases. Each case should have its own independent setup.

internal/connect/cert/issuer.go
clement0010: Yeah there's no signer here, what I meant should be "signed by a different signer's public key"

internal/connect/cert/issuer_test.go
clement0010: Ah, good catch. I think constructing the `x509.Certificate` directly would've avoided the issue, though it'd have more duplicated lines.

internal/connect/cert/issuer_test.go
clement0010: Hmm, I always had the impression that helper test function should declare it. But looking deeper, it's only useful when the function calls `t.Error/t.Fatal/t.Log` internally, so the log is surfaced to the caller instead. 

In this scenario, the function can't fail a test, so we can omit it.

minhtule (approved):
LGTM 👍

PR #450, opened 2026-08-19
fix: sign the SSH host certificate for the requested host and aliases

minhtule:

- closes #442
- stacked on #452, which separates the requested host from the upstream address on `connect.Conn`

## Changes

- Sign the Gateway's host certificate for the host the CONNECT request carried plus the resource's  aliases, since OpenSSH 10.3 rejects a certificate with an empty principal list
  - The resource address is excluded because it may be a wildcard, which older clients do not match consistently. In addition, we might support CIDR address in the future but OpenSSH doesn't support CIDR.
- Move host certificate signing from proxy startup to the handshake, where the requested host is known
  - One cached certificate per principal set, renewed in the background and re-signed on CA key rotation
  - Cache bounded at 1024 entries, evicted after a full certificate lifetime without use

## Notes

- ⚠️ Existing Vault deployments must give the Gateway's host CA role an `allowed_domains` covering the resource addresses; a role relying on `allow_empty_principals` no longer signs these certificates.


internal/sshhandler/cert.go
```diff
@@ -50,33 +63,222 @@ type certificateRequest struct {
...
+func (c *hostCertManager) start(ctx context.Context) {
+	go c.maintenanceLoop(ctx)
+}
+
+// maintenanceLoop runs until ctx is canceled. Keeping a certificate signed by the CA's current key
+// belongs to each cached certificate, not here.
+func (c *hostCertManager) maintenanceLoop(ctx context.Context) {
+	// Sweeping twice per lifetime keeps a certificate that was last used just after a sweep from
+	// renewing for another full lifetime before the next one sees it.
+	ticker := time.NewTicker(c.ttl / 2)
```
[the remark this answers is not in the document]
minhtule: Revert back to just use TTL for simplicity for now.

internal/sshhandler/cert.go
```diff
@@ -50,15 +59,175 @@ type certificateRequest struct {
...
+	return certSigner, nil
+}
+
+func (c *hostCertManager) start(ctx context.Context) {
+	go c.maintenanceLoop(ctx)
+}
+
+// maintenanceLoop runs until ctx is canceled. Renewal belongs to each cached certificate, not here.
+func (c *hostCertManager) maintenanceLoop(ctx context.Context) {
+	ticker := time.NewTicker(c.ttl)
```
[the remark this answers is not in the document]
minhtule: This is good enough for now.

internal/sshhandler/cert_test.go
```diff
@@ -43,182 +67,522 @@ func (c *stubCA) sign(_ context.Context, req *certificateRequest) (*ssh.Certific
...
+}
+
 func (c *stubCA) calls() int {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	return c.signCalls
 }
 
-func TestAutoRenewingCertSigner_RenewsAtExpectedTime(t *testing.T) {
```
minhtule, line 66: This test is moved to `TestHostCertManager_ResignsAtRenewTime` as the renewal behavior could be provided by the host cert manager.

internal/sshhandler/cert_test.go
```diff
@@ -43,182 +67,522 @@ func (c *stubCA) sign(_ context.Context, req *certificateRequest) (*ssh.Certific
...
+		assert.Equal(t, 2, ca.calls(), "a handshake should not wait on the CA to retry the renewal")
+
+		time.Sleep(retryInterval)
+		synctest.Wait()
+		assert.Equal(t, 3, ca.calls(), "the renewal should be retried after retryInterval")
+		assert.NotEqual(t, original, signedCert(t, signer).Marshal())
 	})
 }
 
-func TestAutoRenewingCertSigner_RetriesOnRenewError(t *testing.T) {
```
minhtule, line 128: Moved to `TestHostCertManager_ServesCurrentCertificateWhenSignFails`

internal/sshhandler/proxy_test.go
```diff
@@ -76,59 +76,24 @@ func TestProxy_ProxiesConnection(t *testing.T) {
...
-						logger:     zap.NewNop(),
-					},
-				}
-			},
-			wantErr: "failed to login to Vault",
-		},
-		{
-			name: "downstream config failure",
-			setup: func(t *testing.T, sshProxy *SSHProxy) {
-				t.Helper()
```
minhtule, line 110: The downstream config is constructed per connection, not at the beginning of the SSH proxy setup.

test/integration/testutil/vault.go
```diff
@@ -87,7 +87,7 @@ func SetupVaultServer(t *testing.T) (string, int) {
 		"allow_user_certificates=true",
 		"allow_host_certificates=true",
 		"allowed_extensions=permit-port-forwarding,permit-pty,permit-user-rc",
-		"allow_empty_principals=true",
+		"allowed_domains=*",
```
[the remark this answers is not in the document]
minhtule: This is not true. The integration test passed.

internal/sshhandler/cert.go
```diff
@@ -50,15 +62,213 @@ type certificateRequest struct {
...
+}
+
+// evictAll drops every certificate once the manager stops. Renewal ends with the manager's context,
+// even for a certificate cached after this returns, so there is nothing to cancel here.
+func (c *hostCertManager) evictAll() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	clear(c.certByPrincipals)
+}
```
clement0010, line 202: Maybe we could consider using [golang-lru](https://pkg.go.dev/github.com/hashicorp/golang-lru/v2), so we don't have to write this LRU logic? 🤔 

We don't have to make the change in this PR. I think the dynamic TLS cert signing has similar LRU cache logic. We can refactor this to a shared module or maybe use `golang-lru`

internal/sshhandler/cert.go
```diff
@@ -50,15 +62,213 @@ type certificateRequest struct {
...
+		c.evictOldest()
+	}
+
+	// #nosec G118 -- the cached certificate owns cancelRenewal and calls it when it is replaced or evicted
+	renewalCtx, cancelRenewal := context.WithCancel(c.renewalRoot)
+
+	//nolint:contextcheck // the cache serves this certificate to connections other than the one that signed it
+	go certSigner.renewalLoop(renewalCtx)
+
+	c.certByPrincipals[key] = cachedHostCert{
```
sghiocel, line 160: Somewhat minor since it's during shutdown, but this is a possibility:

Thread A -> serving a connection
Thread M -> maintenance thread
```
A  cert.go:108   c.mu.Lock()
A  cert.go:110   cache miss for this principal set
A  cert.go:123   c.mu.Unlock()          ← lock dropped
A  cert.go:127   newAutoRenewingCertSigner(...)   ← slow: up to 60s against Vault

M  cert.go:179   <-ctx.Done() fires (Gateway shutting down)
M  cert.go:180   c.evictAll()
M  cert.go:192       c.mu.Lock()
M  cert.go:195       clear(c.certByPrincipals)
M  cert.go:193       c.mu.Unlock()  (deferred)
M  cert.go:182   return              ← maintenanceLoop is gone; evictUnused will never run again

A  cert.go:137   c.mu.Lock()          ← A wakes up into a world with no maintenance
A  cert.go:142   map is empty → no `replaced` branch
A  cert.go:144   len(map)==0 → no evictOldest
A  cert.go:149   context.WithCancel(c.renewalRoot)   ← renewalRoot is ALREADY cancelled
A  cert.go:152   go renewalLoop(renewalCtx)
                     → cert.go:341 <-ctx.Done() fires immediately → returns, never renews
A  cert.go:154   c.certByPrincipals[key] = ...        ← inserts into the map evictAll just cleared
```

Can be fixed by checking and not caching in this case:
```
+	// The manager stopped while we were signing. Nothing will renew or evict this entry, so
+	// hand the certificate to this connection without caching it.
+	if c.renewalRoot.Err() != nil {
+		return certSigner, nil
+	}
+
 	if replaced, ok := c.certByPrincipals[key]; ok {
```

sghiocel (approved):
One comment, otherwise looks good. Fix & merge

internal/sshhandler/cert.go
```diff
@@ -50,15 +62,213 @@ type certificateRequest struct {
...
+		signer:        certSigner,
+		cancelRenewal: cancelRenewal,
+		lastUsed:      now,
+	}
+
+	return certSigner, nil
```
clement0010, line 166: nit: maybe we want to add a debug/info log for issuing a new cert successfully?

internal/sshhandler/cert.go
```diff
@@ -116,21 +336,31 @@ func (s *autoRenewingCertSigner) renewalLoop(ctx context.Context) error {
 	for {
 		select {
 		case <-timer.C:
-		case <-s.renewCh:
-			s.logger.Info("CA key reloaded, re-signing certificate", zap.String("cert_type", s.certReq.certType.String()))
+		case <-s.rotated:
+			s.updateRotationChannel()
```
clement0010, line 346: nit: maybe we should keep the debug log so we can troubleshoot when a CA is rotated?

internal/sshhandler/cert_test.go
```diff
@@ -32,193 +41,429 @@ func (c *stubCA) publicKey(_ context.Context) (ssh.PublicKey, error) {
...
+		ca := newStubCA(t)
+		manager := newTestHostCertManager(t, ca)
 
-		keySigner, _, err := keyConfig{}.Generate(rand.Reader)
-		require.NoError(t, err)
+		ctx, shutdown := context.WithCancel(t.Context())
+		manager.start(ctx)
+
+		shutdown()
+		synctest.Wait() // maintenance clears the cache and exits
```
clement0010, line 290: I think we can cache one cert before this, then assert the cache is truly empty.

clement0010 (approved):
LGTM 👍🏻  A few comments

internal/sshhandler/cert_test.go
minhtule: Added in https://github.com/Twingate/gateway/pull/450/commits/1b001205e0dd24d110cfa849b466c93d7a39b7ad

internal/sshhandler/cert.go
minhtule: Added in https://github.com/Twingate/gateway/pull/450/commits/6e2c59f571341df9dda7f6ee2ad3bcc83ebd510c

internal/sshhandler/cert.go
minhtule: Added https://github.com/Twingate/gateway/pull/450/commits/724d4c1a70a1c24b9251dae1b832831ce36f1c6d

internal/sshhandler/cert.go
minhtule: Not sure we want this because we already have a log when the key file changes. The log here is for when the cert gets the rotate signal. There could be 1024 host certs now so having this log seems to be noisy. We also have a log when the cert signing fails or succeeds.

internal/sshhandler/cert.go
```diff
@@ -98,16 +311,29 @@ func (s *autoRenewingCertSigner) Sign(rand io.Reader, data []byte) (*ssh.Signatu
...
 
-	cert := certSigner.PublicKey().(*ssh.Certificate) //revive:disable:unchecked-type-assertion -- certSigner always wraps a certificate
+	return s.expiresAt.IsZero() || t.Before(s.expiresAt)
+}
+
+// updateRotationChannel points s.rotated at the CA's current rotation channel.
+func (s *autoRenewingCertSigner) updateRotationChannel() {
+	if rotatable, ok := s.ca.(rotatableCA); ok {
+		s.rotated = rotatable.rotated()
+	}
```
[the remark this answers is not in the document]
minhtule: this is an entirely new feature...

internal/sshhandler/cert.go
minhtule: The caching mechanism here is a bit more sophisticated than an LRU: because of the background renewal loop in each cert, the cache needs to purge unused one. So it needs to know each cert's last used which is internal in `golang-lru`. Let's see how the TLS cert signing logic in https://github.com/Twingate/gateway/pull/425 works and decide later if we want to create an abstraction.

PR #452, opened 2026-08-20
refactor(connect): store the requested host from the CONNECT message

minhtule:
## Changes

- `connect.Conn` keeps the host from the CONNECT target alongside the upstream host it resolves to;
  previously alias resolution overwrote it and only the upstream address was stored

## Notes

The requested host will be needed for https://github.com/Twingate/gateway/pull/450

### DIFF
--- a/internal/connect/conn.go
+++ b/internal/connect/conn.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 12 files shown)


internal/connect/conn.go
```diff
@@ -33,13 +34,14 @@ func httpResponseString(httpCode int) string {
 }
 
 // Conn is a custom connection that wraps the underlying TCP net.Conn, handling downstream
-// proxy (Twingate Client)'s authentication via the initial CONNECT message. It handles 2 TLS
-// upgrades: with downstream proxy and then optionally with downstream client e.g. `kubectl`.
```
minhtule, line 37: The 2nd sentence is no longer true. This `Conn` should only be responsible for the outer TLS.

sghiocel (approved):

clement0010 (approved):
LGTM 👍🏻

Issue #460, opened 2026-08-25
fix(ssh): sequence channel-request events so the session recorder is in place before they are applied


minhtule:
`SSHRequestHandler` reaches `SSHChannelPair` over two paths that are not ordered against each other: `onPtyRequest` and `onWindowChange` run inline on the request-handler goroutine, while `sessionSignals.started` hands off to `serve()`, which only then creates the session recorder those callbacks read.

The handler replies to the session-start request before signaling `started`, so a client holding that reply can get a `window-change` back to the handler before `serve()` is scheduled. `onWindowChange` then reads a nil recorder and the resize is dropped from the recording. `recMu` makes the shared recorder safe to touch but does not order its creation before a callback that needs it.

Surfaced as a flaky `TestChannelPair_SessionRecorderWriteErrors`; its resize assertion is currently parked behind a TODO in `internal/sshhandler/channel_pair_test.go`.

Desired outcome:

- One ordered path between the request handler and the channel pair, so an event that follows the session start is applied after the recorder exists.
- A single owner of the recorder, retiring the shared variable and `recMu`.
- Restore the parked resize coverage.

PR #461, opened 2026-08-25
test: Fix a randomly failed test

minhtule:

- Issue: https://github.com/Twingate/gateway/issues/460

## Changes

Skip test and comment out `window-change` channel request in flaky tests due to race condition (see https://github.com/Twingate/gateway/issues/460)

### DIFF
--- a/internal/sshhandler/channel_pair_test.go
+++ b/internal/sshhandler/channel_pair_test.go
@@ -722,10 +722,7 @@ func TestChannelPair_CopyPanicClosesChannels(t *testing.T) {
 }
 
 func TestChannelPair_RequestPanicClosesChannels(t *testing.T) {
-	//nolint:godox
-	// TODO(#460): flaky, the window-change can be handled before serve() creates the recorder,
-	// so no resize panic fires and the channel stays open.
-	t.Skip()
+	t.Skip("TODO(#460): flaky; the window-change can be handled before serve() creates the recorder, so no resize panic fires and the channel stays open")
 
 	// A panic in a request handler (here from the recorder's resize write) must tear down the
 	// channel instead of leaving it open with nobody consuming its requests.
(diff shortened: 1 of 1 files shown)


internal/sshhandler/channel_pair_test.go
```diff
@@ -678,12 +678,14 @@ func TestChannelPair_SessionWindowChangeWithoutRecorder(t *testing.T) {
...
-	channels.recorder.resizeErr = errors.New("resize write failed")
+	//nolint:godox
+	// TODO(#460): flaky, the window-change can be handled before serve() creates the recorder.
+	// channels.recorder.resizeErr = errors.New("resize write failed")
 	done := channels.serve(t)
 
 	channels.sendRequestAwaitReply(t, requestTypePty, ssh.Marshal(ptyReq{Term: "xterm", WidthColumns: 80, HeightRows: 24}))
 	channels.sendRequestAwaitReply(t, requestTypeShell, nil)
-	channels.sendWindowChange(t, 120, 40)
+	// channels.sendWindowChange(t, 120, 40)
```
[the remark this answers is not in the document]
minhtule: We'll fix the code later to avoid this race window.

internal/sshhandler/channel_pair_test.go
sghiocel, line 739: I think this is flaky too, because if `rec == nil`, it won't panic and EOF checks here will timeout

internal/sshhandler/channel_pair_test.go
```diff
@@ -678,12 +678,14 @@ func TestChannelPair_SessionWindowChangeWithoutRecorder(t *testing.T) {
 func TestChannelPair_SessionRecorderWriteErrors(t *testing.T) {
 	channels := newProxyChannels(t, "session")
 	channels.recorder.headerErr = errors.New("header write failed")
-	channels.recorder.resizeErr = errors.New("resize write failed")
+	//nolint:godox
+	// TODO(#460): flaky, the window-change can be handled before serve() creates the recorder.
```
sghiocel, line 682: Hm, we could use `t.Skip` to make it cleaner vs commenting out these lines, but it would require splitting these into two tests ... your call

sghiocel (approved):
2 comments, decide + fix + merge

PR #489, opened 2026-09-01
feat(ssh)!: Rename `ssh.ca.manual` config to `ssh.ca.local`

minhtule:
⚠️ This is a schema breaking change for `config.yaml`.

## Changes

Rename "manual CA" to "local CA" for SSH config to be consistent with local issuer for TLS in https://github.com/Twingate/gateway/issues/423


clement0010 (approved):
LGTM 👍

PR #514, opened 2026-09-06
feat(helm): allow cert-manager to use a public ACME issuer

minhtule:
## Changes

- Add the `kubernetes` Service DNS names to the certificate only when the Gateway serves Kubernetes resources (`kubernetes.enabled` or `twingateOperator.kubernetesResource.enabled`), since a public CA refuses them.
  - `kubernetes`, `kubernetes.default`, `kubernetes.default.svc`, `kubernetes.default.svc.<clusterDomain>`
- Emit `tls.autoGenerated.certManager.commonName` only when set, instead of defaulting to the chart fullname, which is equally unissuable by a public CA.
- Fail the render with an actionable message when neither `tls.dnsNames` nor `tls.ipAddresses` leaves the certificate any identifier.

## Notes

- Existing installs on the helm TLS engine (the default) regenerate `tls.crt`, `tls.key` and `ca.crt` on the first upgrade, because the alternative-names hash now covers a different string and no longer matches the `checksum/alternativeNames` annotation on the existing Secret. Releases with `twingateOperator.gateway.enabled: true` re-register the new `ca.crt` through the operator; those that registered the CA by hand have to re-register it.


deploy/gateway/templates/_helpers.tpl
```diff
@@ -203,12 +203,42 @@ Get the alias of the in-cluster Kubernetes resource
...
+{{- $dnsNames = append $dnsNames $resourceAlias }}
+{{- end }}
+{{- $dnsNames | toJson }}
+{{- end }}
+
+{{/*
+Fail when the certificate the chart would issue covers no name.
+*/}}
+{{- define "gateway.requireTlsNames" -}}
+{{- if and (not (include "gateway.tlsDnsNames" . | fromJsonArray)) (not .Values.tls.ipAddresses) }}
```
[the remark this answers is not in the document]
minhtule: Fixed in https://github.com/Twingate/gateway/pull/514/commits/393ca36b87c554f75138e1d04b61e97bb75d1db6

deploy/gateway/values.schema.json
```diff
@@ -451,16 +451,18 @@
         "ipAddresses": {
           "type": "array",
           "items": {
-            "type": "string"
+            "type": "string",
+            "minLength": 1
```
clement0010, line 455: We could add a simple unit test for this:
```yaml
set:
  tls:
    autoGenerated:
      enabled: true
    dnsNames: [""]
```
in `tls-secret_test.yaml`

deploy/gateway/values.schema.json
```diff
@@ -451,16 +451,18 @@
...
           },
           "description": "IP addresses to include as Subject Alternative Names in the TLS certificate."
         },
         "dnsNames": {
           "type": "array",
           "items": {
-            "type": "string"
+            "type": "string",
+            "minLength": 1
```
clement0010, line 463: ditto

clement0010 (approved):
LGTM 👍🏻 

One question: Since public CA can't sign the K8s DNS name, does that mean we can't enable both k8s and web app for this specifc setup?

minhtule:
> Since public CA can't sign the K8s DNS name, does that mean we can't enable both k8s and web app for this specifc setup?

No, not until we support multiple CAs per Gateway in the backend.

deploy/gateway/values.schema.json
minhtule: Added in https://github.com/Twingate/gateway/pull/514/commits/be662e6877399e9ac62aba103a6120c02ea7a228

deploy/gateway/values.schema.json
minhtule: https://github.com/Twingate/gateway/pull/514/commits/be662e6877399e9ac62aba103a6120c02ea7a228

PR #522, opened 2026-09-08
feat: support GCP Private CA as TLS automation cert issuer

clement0010:
## Related Tickets

Issue: #423

## Changes

- Support `tls.automation.issuer.gcpPrivateCA` config. When configured, Gateway issues downstream TLS certs through a Google Cloud CA Service pool, and verifies each issued cert the same way the Vault issuer does: exact requested DNS/IP SANs, TTL and public key.
- Credentials come from a service account key file, or from Application Default Credentials when none is configured.


internal/config/config.go
```diff
@@ -558,6 +582,22 @@ func (v *TLSVaultIssuerConfig) Validate() error {
 	return nil
 }
 
+func (g *TLSGCPPrivateCAIssuerConfig) Validate() error {
+	if g.Project == "" {
+		return fmt.Errorf("%w: project", ErrRequired)
+	}
```
clement0010, line 588: I think it's simpler to make `project` field required so we don't have to make a separate API call to query the project name. 

[cert-manager](https://github.com/cert-manager/google-cas-issuer#configuring-the-issuer) requires it as well

internal/connect/cert/issuer.go
```diff
@@ -257,6 +265,78 @@ func (v *vaultIssuer) sign(ctx context.Context, req *certificateRequest) (*x509.
...
+}
+
+// run opens the CA Service client, which then lives for the process lifetime.
+func (g *gcpIssuer) run(ctx context.Context) error {
+	var opts []option.ClientOption
+	if g.credentialsFile != "" {
+		opts = append(opts, option.WithAuthCredentialsFile(option.ServiceAccount, g.credentialsFile))
+	}
+
+	client, err := privateca.NewCertificateAuthorityClient(ctx, opts...)
```
clement0010, line 306: The default client is `gRPC`-based

If we want to be consistent with Vault issuer, there is also a REST-based client  [`NewCertificateAuthorityRESTClient()`](https://pkg.go.dev/cloud.google.com/go/security/privateca/apiv1#NewCertificateAuthorityRESTClient).

internal/connect/cert/issuer.go
```diff
@@ -257,6 +265,78 @@ func (v *vaultIssuer) sign(ctx context.Context, req *certificateRequest) (*x509.
...
+		issuingCA:       cfg.IssuingCertificateAuthorityID,
+	}
+}
+
+// run opens the CA Service client, which then lives for the process lifetime.
+func (g *gcpIssuer) run(ctx context.Context) error {
+	var opts []option.ClientOption
+	if g.credentialsFile != "" {
+		opts = append(opts, option.WithAuthCredentialsFile(option.ServiceAccount, g.credentialsFile))
+	}
```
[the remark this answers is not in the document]
clement0010: This is not true. The option does exist

internal/connect/cert/issuer.go
```diff
@@ -257,6 +265,86 @@ func (v *vaultIssuer) sign(ctx context.Context, req *certificateRequest) (*x509.
 	return leaf, chain[1:], nil
 }
 
+// gcpIssuer issues leaf certificates through Google Cloud Certificate Authority Service.
+type gcpIssuer struct {
```
minhtule: I would name this `gcpPrivateCA` because in the future, we might support signing with GCP Public CA or using GCP KMS.

minhtule (approved):
LGTM 👍

internal/connect/cert/issuer.go
```diff
@@ -271,6 +279,85 @@ func (v *vaultIssuer) sign(ctx context.Context, req *certificateRequest) (*x509.
...
+			Lifetime: durationpb.New(req.ttl),
+		},
+		RequestId:                     uuid.NewString(),
+		IssuingCertificateAuthorityId: g.issuingCA,
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to issue certificate: %w", err)
+	}
+
+	chain, err := parseCertificateChain(append([]string{issued.GetPemCertificate()}, issued.GetPemCertificateChain()...))
```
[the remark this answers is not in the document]
clement0010: The chain starts at the CA that signed the leaf

PR #523, opened 2026-09-08
fix: support Host header rewrite

minhtule:
## Changes

- `Host` in `webApp.requestHeaders` or a GAT request header rewrite now replaces the Host header sent to the upstream
  - `net/http` ignores a `Host` entry in the header map and sends `Request.Host` instead, so the rewrite was silently dropped

### DIFF
--- a/internal/webapphandler/handler.go
+++ b/internal/webapphandler/handler.go
@@ -118,7 +118,7 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
 			return fmt.Errorf("header %q: %w", headerName, err)
 		}
 
-		r.Out.Header.Set(headerName, headerValue)
+		setHeader(r, headerName, headerValue)
 	}
 
 	// Per-resource request header rewrites from the GAT are applied last, so they override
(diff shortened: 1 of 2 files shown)


clement0010 (approved):
LGTM 👍🏻

internal/webapphandler/handler.go
```diff
@@ -139,8 +139,17 @@ func rewrite(r *httputil.ProxyRequest, conn *connect.ProxyConn, headers map[stri
...
+		setHeader(r, headerName, headerValue)
 	}
 
 	return nil
 }
+
+func setHeader(r *httputil.ProxyRequest, name, value string) {
+	switch http.CanonicalHeaderKey(name) {
+	case "Host":
+		r.Out.Host = value
```
sghiocel, line 151: We have the following comment on line 124-126:
```
	// Per-resource request header rewrites from the GAT are applied last, so they override
	// any config headers with the same name. Malformed or unsupported headers are skipped
	// rather than failing the request.
```

So for GAT rewrites, we expect to proceed with the request if the header is malformed. From what I understand though the `Host` header is under additional validation:

HTTP/1.1: https://github.com/golang/go/blob/go1.27.0/src/net/http/request.go#L631
(changes it into an empty string "" if it does not validate)

HTTP/2:
https://github.com/golang/go/blob/go1.27.0/src/net/http/internal/httpcommon/httpcommon.go#L235-L237 (fails the request with `502`)

So the question is, do we want to preserve this behavior? Do we want to validate the `Host` header ourselves up front and for Config rewrites, have our own error, and for GAT try to skip as we did before? Or are we OK with empty Host or `502` in this invalid cases (if it includes invalid symbols like `@` `/` `#` etc)

minhtule: I think the comment is a bit misleading. We only skip the header if the template fails to parse or evaluate (updated in https://github.com/Twingate/gateway/pull/523/commits/7295c226925bc533ca06abc7d13e0d0b7a3f2078). We never validate the header name and value according to the RFCs. Go would do those validations and fail the request instead.

sghiocel (approved):

